PR #5, opened 2021-05-14
Hw dalai

Dalai-S:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,65 @@
...
+
     public static void main(String[] args) {
-        // test your code here
+        Manufacturer audi = new Manufacturer("AUDI", "Germany");
+        manufacturerService.create(audi);
+
+        Driver yevheniiParchevskiy = new Driver("Yevhenii Parchevskiy", "11111111111111111");
+        Driver alexanderKabanov = new Driver("Alexander Kabanov", "22222222222222");
+        Driver illyaNemykin = new Driver("Illya Nemykin", "33333333333");
+        Driver arianaGrande = new Driver("Ariana Grande", "444444444444");
```
freedmanMain, line 29: LOL, nice uberdriver =)))

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't create car "
+                    + car + ". ", throwable);
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT cars.id as car_id, manufacturers.id as manufacturer_id, "
+                + "manufacturers.name, manufacturers.country "
```
freedmanMain, line 43: You can make it easier. Use 'SQL-Alias' it is best practice.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+                driverList.add(getDriverFromResultSet(resultSet));
+            }
+            return driverList;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Cannot get drivers for car id: " + id, e);
+        }
+    }
+
+    private void addDriversForCars(List<Car> cars) {
+        if (cars.size() > 0) {
```
freedmanMain, line 202: I think it would be better to add a null check here. Because if the size of the list is zero, we won't even get into this loop. But if it is null, we can get NPE.

freedmanMain (approved):
good job.

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,68 @@
+package mate.jdbc.model;
+
+import java.util.List;
+import java.util.Objects;
+
+public class Car {
+    private Long id;
+    private Manufacturer manufacturer;
```
Max-Pochepets, line 8:
```suggestion
    private Long id;
    private String model;
    private Manufacturer manufacturer;
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query,
+                         Statement.RETURN_GENERATED_KEYS)) {
+            statement.setLong(1, car.getManufacturer().getId());
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            addDriversToDb(car);
```
Max-Pochepets, line 32: you're opening connection inside a connection. let's keep only one connection opened at a time

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        car.setManufacturer(getManufacturer(resultSet.getObject("manufacturer_id", Long.class)));
+        car.setId(resultSet.getObject("car_id", Long.class));
+        return car;
+    }
+
+    private Manufacturer getManufacturer(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = "
```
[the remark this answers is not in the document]
Dalai-S: Because here I create a car object and I have no information about manufacturer in "cars" table and have only in "manufacturers" table. In get qurey for car I can get only id, which is used to find other info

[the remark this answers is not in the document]
Max-Pochepets: you're passing result set that contains all this information. what is `m.id`, `m.name` and `m.country` if not the info you seek. you can also add to this query m.is_deleted = FALSE
image
you can create new Manufacturer object in `private getCar()`, get all you need from result set and set the car with this object

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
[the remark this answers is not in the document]
Dalai-S: Ok, got it

Max-Pochepets (approved):

PR #9, opened 2021-05-15
added solution

safordog:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private void saveDrivers(Car car) throws SQLException {
+        String query = "INSERT cars_drivers (car_id, driver_id) VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement insertStatement = connection.prepareStatement(query)) {
+            insertStatement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                insertStatement.setLong(2, driver.getId());
+                insertStatement.executeUpdate();
+            }
```
mankokolya, line 182: It is better to use catch here then you will see the exception with descriptive message that error happened when saving drivers but not trying to save the car.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement insertStatement =
+                        connection.prepareStatement(query, Statement.RETURN_GENERATED_KEYS)) {
+            insertStatement.setString(1, car.getModel());
+            insertStatement.setLong(2, car.getManufacturer().getId());
+            insertStatement.executeUpdate();
+            ResultSet resultSet = insertStatement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+                saveDrivers(car);
```
mankokolya, line 32: it is better to save cars after catch block not to keep one extra connection open

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+                + "FROM cars c JOIN manufacturers m ON c.manufacturer_id = m.id "
+                + "WHERE c.id = ? AND c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getByIdStatement = connection.prepareStatement(query)) {
+            getByIdStatement.setLong(1, id);
+            ResultSet resultSet = getByIdStatement.executeQuery();
+            Car car = null;
+            if (resultSet.next()) {
+                car = getCarFromResultSet(resultSet);
+                car.setDrivers(getDriversForCar(id));
```
mankokolya, line 53: Same here setDrivers afters catch block

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+                + "FROM cars c JOIN manufacturers m on c.manufacturer_id = m.id "
+                + "WHERE c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement getAllStatement = connection.createStatement()) {
+            ResultSet resultSet = getAllStatement.executeQuery(query);
+            List<Car> cars = new ArrayList<>();
+            Car car;
+            while (resultSet.next()) {
+                car = getCarFromResultSet(resultSet);
+                car.setDrivers(getDriversForCar(car.getId()));
```
mankokolya, line 73: Set drivers after catch block

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT c.id, model, manufacturer_id, name, country "
+                + "FROM cars c JOIN manufacturers m on c.manufacturer_id = m.id "
+                + "WHERE c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement getAllStatement = connection.createStatement()) {
+            ResultSet resultSet = getAllStatement.executeQuery(query);
+            List<Car> cars = new ArrayList<>();
+            Car car;
```
mankokolya, line 70: no need in this car

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        String query = "UPDATE cars SET model = ?, manufacturer_id = ?"
+                + " WHERE id = ? AND is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement updateStatement = connection.prepareStatement(query)) {
+            updateStatement.setString(1, car.getModel());
+            updateStatement.setLong(2, car.getManufacturer().getId());
+            updateStatement.setLong(3, car.getId());
+            updateStatement.executeUpdate();
+            deleteCarDrivers(car.getId());
+            saveDrivers(car);
```
mankokolya, line 94: after catch block

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        String query = "SELECT car_id FROM cars_drivers c_d JOIN cars c"
+                + " ON c.id = c_d.car_id"
+                + " WHERE driver_id = ? AND c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllStatement = connection.prepareStatement(query)) {
+            getAllStatement.setLong(1, driverId);
+            ResultSet resultSet = getAllStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                cars.add(get(resultSet.getObject(1, Long.class)).orElseThrow(() ->
```
mankokolya, line 125: I don't understand this get() method where it is from? Where do you initialize your car with manufacturer and drivers

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    private Car getCarFromResultSet(ResultSet resultSet) throws SQLException {
+        Long manufacturerId = resultSet.getObject("manufacturer_id", Long.class);
+        String manufacturerName = resultSet.getString("name");
+        String manufacturerCountry = resultSet.getString("country");
+        Manufacturer manufacturer = new Manufacturer(manufacturerName, manufacturerCountry);
+        manufacturer.setId(manufacturerId);
+        Long carId = resultSet.getObject(1, Long.class);
```
mankokolya, line 142: use column name instead of integer

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        String carModel = resultSet.getString("model");
+        Car car = new Car(carModel, manufacturer);
+        car.setId(carId);
+        return car;
+    }
+
+    private List<Driver> getDriversForCar(Long carId) {
```
mankokolya, line 149: getDriversByCarId is better naming as you pass as parameter only id

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,84 @@
...
+        this.model = model;
+        this.manufacturer = manufacturer;
+    }
+
+    public Car(String model, Manufacturer manufacturer, List<Driver> drivers) {
+        this.model = model;
+        this.manufacturer = manufacturer;
+        this.drivers = drivers;
+    }
+
```
mankokolya, line 22: no need to pass drivers through constructor. As you don't need them all the time. just initialize drivers = new ArrayList in the constructor above. And use setter for initializing

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE `manufacturers` (
```
mankokolya, line 1: Don't see create instruction for database creation

mankokolya (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
safordog: Тут я використовував метод get(carId). Змінив все. Дякую!

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -29,28 +29,29 @@ public Car create(Car car) {
             ResultSet resultSet = insertStatement.getGeneratedKeys();
             if (resultSet.next()) {
                 car.setId(resultSet.getObject(1, Long.class));
-                saveDrivers(car);
```
mankokolya: create some boolean helper variable and set it to true in the if body if the resultset wasn't empty

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -29,28 +29,29 @@ public Car create(Car car) {
...
             }
-            return car;
         } catch (SQLException throwable) {
             throw new DataProcessingException("Can't create car: "
                     + car + ". ", throwable);
         }
```
mankokolya, line 36: check if(helper variable is true and only then saveDrivers)

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -29,28 +29,29 @@ public Car create(Car car) {
...
+        saveDrivers(car);
+        return car;
     }
 
     @Override
     public Optional<Car> get(Long id) {
         String query = "SELECT c.id, model, manufacturer_id, name, country "
                 + "FROM cars c JOIN manufacturers m ON c.manufacturer_id = m.id "
                 + "WHERE c.id = ? AND c.is_deleted = FALSE";
+        List<Driver> drivers = getDriversByCarId(id);
```
mankokolya, line 46: You have to get car from db first and if it present only then look for drivers

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -29,28 +29,29 @@ public Car create(Car car) {
...
         try (Connection connection = ConnectionUtil.getConnection();
                 PreparedStatement getByIdStatement = connection.prepareStatement(query)) {
             getByIdStatement.setLong(1, id);
             ResultSet resultSet = getByIdStatement.executeQuery();
             Car car = null;
             if (resultSet.next()) {
                 car = getCarFromResultSet(resultSet);
-                car.setDrivers(getDriversForCar(id));
+                car.setDrivers(drivers);
             }
```
mankokolya, line 55: check after catch if(car != null) {
search for drivers}
return car

mankokolya: like in getAll

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -63,21 +64,21 @@ public Car create(Car car) {
...
+                cars.add(getCarFromResultSet(resultSet));
             }
-            return cars;
         } catch (SQLException throwable) {
             throw new DataProcessingException("Can't get a list of cars from DB.",
                     throwable);
         }
+        for (Car car : cars) {
+            car.setDrivers(getDriversByCarId(car.getId()));
+        }
```
mankokolya, line 80: try to use stream cars.foreach

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -90,12 +91,12 @@ public Car update(Car car) {
...
             updateStatement.setLong(3, car.getId());
             updateStatement.executeUpdate();
-            deleteCarDrivers(car.getId());
-            saveDrivers(car);
         } catch (SQLException throwable) {
             throw new DataProcessingException("Can't update "
                     + car + " in DB.", throwable);
         }
+        deleteCarDrivers(car.getId());
+        saveDrivers(car);
```
mankokolya, line 99: SAme here add helper boolean or int variable and assign it value of updateStatement.executeUpdate();
then check if the was deleted the delete car and saveDrivers

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+            while (resultSet.next()) {
+                cars.add(getCarFromResultSet(resultSet));
+            }
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Can't get all cars for driver with id: "
+                    + driverId, throwable);
+        }
+        for (Car car : cars) {
+            car.setDrivers(getDriversByCarId(car.getId()));
+        }
```
mankokolya, line 135: use cars.foreach

mankokolya (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
safordog: Дякую!

mankokolya (approved):

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -15,6 +15,12 @@ public Driver(String name, String licenseNumber) {
         this.licenseNumber = licenseNumber;
     }
 
+    public Driver(Long id, String name, String licenseNumber) {
```
Max-Pochepets, line 18: you're getting close to start a Builder) use the default constructor

Max-Pochepets (approved):

PR #10, opened 2021-05-16
Hw

kupora3:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,20 @@
...
+public interface CarDao {
+
+    Car create(Car car);
+
+    Optional<Car> get(Long id);
+
+    List<Car> getAll();
+
+    List<Car> getAllByDriver(Long driverId);
+
```
LayRipper, line 16: this method left unchecked

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+            ResultSet resultSet = getAllDriversStatement.executeQuery();
+            while (resultSet.next()) {
+                Car car = carParser(resultSet);
+                cars.add(car);
+            }
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't get a list of cars from carsDB.",
+                    throwable);
+        }
+        setDriversToCars(cars);
```
LayRipper, line 84: this method could be inside carParser method being part of it since setDtiversTo is never used separately

LayRipper (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
kupora3: this is not the case because the carParser method is also called in the get method, where there is no Cars collection

LayRipper (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private void setDriversToCars(List<Car> cars) {
+        if (cars.size() != 0) {
+            for (Car car : cars) {
+                car.setDriver(getDriversForCars(car.getId()));
+            }
+        }
+    }
+}
```
LayRipper, line 244: this method is redundant

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        car.setId(resultSet.getObject("id",Long.class));
+        return car;
+    }
+
+    private List<Driver> getDriversForCars(Long id) {
+        String getAllDriversRequest = "SELECT drivers.id,name,license_number "
+                + "FROM drivers JOIN cars_drivers "
+                + "ON drivers.id = cars_drivers.driver_id "
+                + "WHERE cars_drivers.car_id = ?";
+        List<Driver> getAllDrivers = new ArrayList<>();
```
LayRipper, line 168:
```suggestion
                + "WHERE cars_drivers.car_id = ?";
        List<Driver> allDriversList = new ArrayList<>();
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    private Driver parseDrivers(ResultSet resultSet) throws SQLException {
+        Driver driver = new Driver();
+        driver.setId(resultSet.getObject("id", Long.class));
+        driver.setLicenseNumber(resultSet.getString("license_number"));
+        driver.setName(resultSet.getString("name"));
+        return driver;
+    }
```
LayRipper, line 189:
```suggestion
    private Driver parseDrivers(ResultSet resultSet) throws SQLException {
        Driver driver = new Driver();
        driver.setId(resultSet.getObject("id", Long.class));
        driver.setLicenseNumber(resultSet.getString("license_number"));
        driver.setName(resultSet.getString("name"));   
        return driver;
        driver.setDrivers(getDriversForCars(driver.getId()));
    }
```

LayRipper (changes requested):

LayRipper (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+            throw new DataProcessingException("Couldn't create car: "
+                    + car, throwable);
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String getRequest = "SELECT cars.id,manufacturers.name, model,country "
```
Max-Pochepets, line 44:
```suggestion
        String getRequest = "SELECT cars.id, manufacturers.name, model, country "
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        setDriversToCars(cars);
+        return cars;
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String getAllByDriverRequest = "SELECT * FROM cars_drivers "
+                                + "JOIN cars ON cars_drivers.car_id = cars.id JOIN manufacturers "
+                                + "ON cars.manufacturer_id = manufacturers.id "
+                                + "WHERE driver_id = ? AND cars.is_deleted = false;";
```
Max-Pochepets, line 93: what if we pass ID of deleted driver?

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
kupora3: I would say that this is not possible, but I revised the code and realized that I have nowhere to check that the driver is not null. So I added this check to the method

Max-Pochepets (approved):

PR #25, opened 2021-05-18
JOIN Solution

vuyko23:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,77 @@
...
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector INJECTOR = Injector.getInstance("mate.jdbc");
+    private static final ManufacturerService MANUFACTURER_SERVICE = (ManufacturerService)
+            INJECTOR.getInstance(ManufacturerService.class);
+    private static final DriverService DRIVER_SERVICE = (DriverService)
+            INJECTOR.getInstance(DriverService.class);
+    private static final CarService CAR_SERVICE = (CarService)
+            INJECTOR.getInstance(CarService.class);
```
Den4eg007, line 20: These classes are not deeply immutable you should name these constants as you would any other variable, meaning don't use UPPER_CASE, please correct me if I'm wrong

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+            throw new DataProcessingException("Can`t get List of all drivers by id " + driverId, e);
+        }
+        for (Car car: cars) {
+            car.setManufacturer(getManufacture(car.getId()));
+            car.setDrivers(getAllDriversList(car.getId()));
+        }
+        return cars;
+    }
+
+    private Car createCar(ResultSet resultSet) throws SQLException {
```
Den4eg007, line 149: Why don't you name it parseCarWithManufacturer and create car with manufacturer

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        Car car = new Car();
+        car.setId(resultSet.getObject("id", Long.class));
+        car.setModel(resultSet.getString("model"));
+        return car;
+    }
+
+    private Manufacturer getManufacture(Long id) {
```
Den4eg007, line 156:
```suggestion
    private Manufacturer getManufacturer(Long id) {
```

Den4eg007, line 156: you don't need to make any queries here, you do one query in the get() method and you get ResultSet you it instead
but I would recommend moving it to createCar() method

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+            while (resultSet.next()) {
+                drivers.add(createDriver(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get List of all Drivers by id " + id, e);
+        }
+        return drivers;
+    }
+
+    private Manufacturer createManufacturer(ResultSet resultSet) throws SQLException {
```
Den4eg007, line 196: why do you need getManufacturer method ?
can you make it simpler ?

Den4eg007 (changes requested):

Den4eg007 (approved):

Max-Pochepets (approved):

PR #31, opened 2021-05-18
First solution

Ihor-Nykyforov:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+            updateCarStatement.setLong(1, car.getId());
+            updateCarStatement.executeUpdate();
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't delete drivers from car id "
+                    + car.getId(), throwable);
+        }
+    }
+
+    private void setDrivers(List<Car> carList) {
+        if (carList != null) {
```
fotomassimo, line 210: it's probably redundant, because we always initializing  List<Car> cars = new ArrayList<>() before method call

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+        String query = "INSERT INTO cars_drivers (car_id, driver_id) VALUE (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement insertDriversStatement
+                        = connection.prepareStatement(query)) {
+            insertDriversStatement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                insertDriversStatement.setLong(2, driver.getId());
+                insertDriversStatement.executeUpdate();
+            }
+            return insertDriversStatement.executeUpdate() > 0;
```
fotomassimo, line 190: It seems like in return statement the  executeUpdate() method is called again and executes the same query again. may be its better to put counter inside the loop and return counter == car.getDrivers().size()?

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+                + "WHERE cd.driver_id = ? AND c.deleted = FALSE;";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllByDriverStatement
+                        = connection.prepareStatement(query)) {
+            getAllByDriverStatement.setLong(1, driverId);
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            while (resultSet.next()) {
+                Car car = getCar(resultSet);
+                cars.add(car);
```
fotomassimo, line 133: same suggestion

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+                + "FROM cars c JOIN manufacturers m on c.manufacturer_id = m.id "
+                + "WHERE c.deleted = FALSE;";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllCarStatement
+                        = connection.prepareStatement(query)) {
+            ResultSet resultSet = getAllCarStatement.executeQuery();
+            while (resultSet.next()) {
+                Car car = getCar(resultSet);
+                cars.add(car);
```
fotomassimo, line 77: suggestion:  cars.add(getCar(resultSet))  instead of creating variable car

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getCarStatement
+                        = connection.prepareStatement(query)) {
+            getCarStatement.setLong(1, id);
+            ResultSet resultSet = getCarStatement.executeQuery();
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't get driver by id " + id, throwable);
```
fotomassimo, line 57: shouldn't be a massage about car?

fotomassimo (changes requested):

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).get();
```
fotomassimo, line 22: suggestion from video lesson about Optional: it is recommended to use orElseThrow() instead of get()

fotomassimo (approved):

Max-Pochepets (approved):

PR #37, opened 2021-05-19
first commit

latkova:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
iedenysenko, line 9: on DAO layer method get must return Optional (as well as DriverDao and ManufacturerDao)

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,211 @@
...
+            ResultSet generatedKeys = createCarStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1, Long.class);
+                car.setId(id);
+            }
+        } catch (SQLException throwables) {
+            throw new DataProcessingException("Can't create a car " + car + " ", throwables);
+        }
+        insertDrivers(car);
+        getDriversForCar(car.getId());
```
iedenysenko, line 38: Why call this method?

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,211 @@
...
+        car.setId(resultSet.getObject("car_id", Long.class));
+        return car;
+    }
+
+    private void insertDrivers(Car car) {
+        String insertDriversRequest = "INSERT INTO cars_drivers (car_id, driver_id) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement addDriverToCarStatement =
+                        connection.prepareStatement(insertDriversRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
```
iedenysenko, line 160: You don't use Statement.RETURN_GENERATED_KEYS anywhere on this method :)

iedenysenko (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
latkova: done

src/main/java/mate/jdbc/dao/CarDaoImpl.java
latkova: removed

src/main/java/mate/jdbc/dao/CarDao.java
latkova: this is neverending point of arguing about returning Object or Optional in this case.
last time I vice versa was critised for returning Optional instead of Object, because in README was stated - return object.
Let stay as it is.
I think the both cases are acceptable.
or we should clear out it by Bogdan.

iedenysenko: ok :)

iedenysenko (approved):

src/main/java/mate/jdbc/dao/CarDao.java
RomanDubovskyi, line 9: It's better to return optional here, because Your differenet services may process the case when obejct is absent differently. That's the default behaviour, that is implemented the same way in Spring Data. 
Besides returning null is a bad practice

RomanDubovskyi (changes requested):

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -33,6 +33,6 @@ public Manufacturer update(Manufacturer manufacturer) {
 
     @Override
     public boolean delete(Long id) {
-        return manufacturerDao.delete(id);
+        return false;
```
RomanDubovskyi, line 36: why did you delete impl?

RomanDubovskyi (changes requested):
All good, but broke ManufacturerService delete method, fix it

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
latkova: maybe accidentally
fixed

RomanDubovskyi (approved):

PR #42, opened 2021-05-19
added solution

PepsiTheDoge:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private void removeDriverFromCar(Car car) {
+        String query = "DELETE FROM cars_drivers WHERE car_id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement removeDriverStatement =
+                        connection.prepareStatement(query)) {
+            removeDriverStatement.setLong(1, car.getId());
+            removeDriverStatement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't remove driver for car: " + car, e);
```
Den4eg007, line 215:
```suggestion
            throw new DataProcessingException("Can't remove driver from car: " + car, e);
```

Den4eg007 (changes requested):

Den4eg007 (approved):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                + "ON c.manufacturer_id = m.id "
+                + "WHERE c.deleted = false;";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllStatement = connection
+                        .prepareStatement(query)) {
+            ResultSet resultSet = getAllStatement.executeQuery();
+            while (resultSet.next()) {
+                Car car = getCarWithManufacturer(resultSet);
+                car.setDrivers(getDriversForCar(car.getId()));
```
Max-Pochepets, line 82: `getDriversForCar` opens connection inside other connection. let's keep one connection opened at a time

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -74,7 +74,7 @@ public Driver create(Driver driver) {
     @Override
     public Driver update(Driver driver) {
         String query = "UPDATE drivers "
-                + "SET name = ?, license_number = ? "
+                + "SET name = ?, license = ? "
```
Max-Pochepets, line 77: no need to change this. class field is `licenseNumber`, two words, so column should also be two words

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
PepsiTheDoge: does it apply to other methods? like addDriverToCar or removeDriverFromCar?

Max-Pochepets: as I see these methods don't contain nested connections. and they don't cause it anymore in this code. GJ

Max-Pochepets (approved):

PR #73, opened 2021-07-16
Homework

Mzuha:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,221 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteByIdRequest = "UPDATE cars SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteByIdStatement =
+                        connection.prepareStatement(deleteByIdRequest)) {
+            deleteByIdStatement.setLong(1, id);
+            int updatedRows = deleteByIdStatement.executeUpdate();
+            return updatedRows >= 0;
```
Tamara-Paskalova, line 114: Here you return true even number of deleted cars is 0. do you think this behavior is OK?

Tamara-Paskalova (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,217 @@
...
+                car = getCarFromResultSet(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Cant get car by id " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getAllDriversByCar(car));
+            return Optional.of(car);
+        }
+        return Optional.empty();
```
Sofasmile, line 64:
```suggestion
        }
        return Optional.ofNullable(car);
```

Sofasmile (approved):

PR #97, opened 2021-10-01
1.0 Solution "jv-join"

Chupiak:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't create " + car + ". ", throwable);
+        }
+    }
+
+    @Override
+    public Car get(Long id) {
+        String query = "SELECT "
+                + "taxi.cars.id AS car_id, "
```
RomanDubovskyi, line 57: do we need schema name before table name here?

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,245 @@
...
+        deleteDrivers(car.getId());
+        createCarsAndDrivers(car);
+        return car;
+
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT "
+                + "taxi.drivers.id AS driver_id, "
```
RomanDubovskyi, line 198: same here, let's shorten this request to improve readability, also IMHO we don't need aliases here

RomanDubovskyi (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Chupiak: removed the maximum number of characters

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,243 @@
...
+        }
+        createCarsAndDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Car get(Long id) {
+        String query = "SELECT "
+                + "cars.id, "
+                + "cars.model, "
```
RomanDubovskyi, line 43:
```suggestion
        String query = "SELECT cars.id, cars.model, "
```
ideally I would form queries like this

RomanDubovskyi (approved):

PR #104, opened 2021-10-04
Hw20 taxi join solution

Yehor-Zapara:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                + "WHERE id = ? AND is_deleted = FALSE;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateCarStatement = connection.prepareStatement(query)) {
+            updateCarStatement.setString(1, car.getModel());
+            updateCarStatement.setLong(2, car.getManufacturer().getId());
+            updateCarStatement.setLong(3, car.getId());
+            updateCarStatement.executeUpdate();
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't update a car "
+                    + car + " ", throwable);
```
Koriastras, line 111:
```suggestion
                    + car + ". ", throwable);
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement insertDriversStatement = connection.prepareStatement(query)) {
+            insertDriversStatement.setLong(1, car.getId());
+            for (Driver driver: car.getDrivers()) {
+                insertDriversStatement.setLong(2, driver.getId());
+                insertDriversStatement.executeUpdate();
+            }
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't insert drivers to car "
+                    + car + " ", throwable);
```
Koriastras, line 168:
```suggestion
                    + car + ". ", throwable);
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,59 @@
...
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        System.out.println("All cars:");
+        List<Car> cars = carService.getAll();
+        cars.forEach(System.out::println);
+        System.out.println(System.lineSeparator() + "Car with id 2:");
+        Car car = carService.get(2L);
+        System.out.println(car);
+        System.out.println(System.lineSeparator() + "Add new car:");
+        Manufacturer manufacturer = new Manufacturer("Peugeot", "France");
+        manufacturer.setId(25L);
```
Koriastras, line 25: we don't need to set id, because the method create() in manufacturerServiceImpl should generate a key

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,59 @@
...
+        drivers.add(driver2);
+        car.setDrivers(drivers);
+        System.out.println(carService.create(car));
+        System.out.println(System.lineSeparator() + "Delete car with id 4:");
+        System.out.println(carService.delete(4L));
+        System.out.println(System.lineSeparator() + "Update car with id 3:");
+        manufacturer = new Manufacturer("Lexus", "Japan");
+        manufacturer.setId(5L);
+        Car car2 = new Car("NX", manufacturer);
+        car2.setId(3L);
```
Koriastras, line 42: the same here, I suppose it would be better to use create() method to create a car and then just set a new value to some of its field and update the data

Koriastras (changes requested):

Koriastras (approved):

PR #108, opened 2021-10-05
Jv join solution

VladGural:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+            ResultSet resultSet = insertCarStatement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject("id", Long.class));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert car"
+                    + car + "  to DB", e);
+        }
+        if (car.getDrivers() != null && car.getDrivers().size() > 0) {
+            setDriversForCar(car.getId(), car.getDrivers());
```
Yelinskyi, line 39: Note: I think would be better just give the car in this method.

Yelinskyi (approved):

Resci (approved):

PR #112, opened 2021-10-05
Created jv-join project

18Yuliya18:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+                        = connection.prepareStatement(query)) {
+            updateCarStatement.setString(1, car.getModel());
+            updateCarStatement.setLong(2, car.getManufacturer().getId());
+            updateCarStatement.setLong(3, car.getId());
+            updateCarStatement.executeUpdate();
+            return car;
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't update "
+                    + car + " in cars DB.", throwable);
+        }
```
SolnalarmJulia, line 107: We have to:
 1) delete all existing relationships between this car and its old drivers;
 2) create new relationships between car and its new driver list.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+    private Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
+        String country = resultSet.getString("country");
+        Manufacturer manufacturer = new Manufacturer(name, country);
+        manufacturer.setId(id);
+        return manufacturer;
+    }
+
+    public List<Driver> getDriversList(Long id) {
```
SolnalarmJulia, line 168: It would be better if name of method is something like getDriversListByCar

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        car.getDrivers().add(driver);
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
```
SolnalarmJulia, line 50: I think we need to check car is exist or not inDB, because we have NullPointerException if car is not exist in DB

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,230 @@
...
+                addDriverStatement.setLong(2, driver.getId());
+                addDriverStatement.executeUpdate();
+            }
+        } catch (SQLException throwables) {
+            throw new DataProcessingException("Couldn't add driver to the cars_drivers "
+                    + car, throwables);
+        }
+    }
+
+    private void deleteCarByCarId(Long carId) {
```
Resci, line 217:
```suggestion
    private void deleteAllDriversByCarId(Long carId) {
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,230 @@
...
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
+        String licenseNumber = resultSet.getString("license_number");
+        Driver driver = new Driver(name, licenseNumber);
+        driver.setId(id);
+        return driver;
+    }
+
+    private void addDriverToCar(Car car) {
```
Resci, line 201:
```suggestion
    private void addDriversToCar(Car car) {
```

Resci (approved):

PR #159, opened 2021-12-06
implemented task jv-join

Nata-Vidzivashets:


SarumanIsAlive (approved):

src/main/resources/init_db.sql
```diff
@@ -1,20 +1,41 @@
-CREATE SCHEMA IF NOT EXISTS `taxi` DEFAULT CHARACTER SET utf8;
-USE `taxi`;
+CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
+USE `taxi_service`;
 
 DROP TABLE IF EXISTS `manufacturers`;
```
issugy, line 4: this wont work if there is any data in ur tables as far as u will have rows referenced on ones from this table
same below

issugy (approved):

PR #179, opened 2021-12-12
Wrote an assignment jv-join.

ASelivonchyk:
- created class Car in package model;
- created CarDao with CRUD methods and implement it in class CarDaoImpl;
- created CarService and implement it in class CarServiceImpl;
- added to init_db.sql parameters to create "cars" and "car_drivers" tables in DB;
in Main class call all of CarsService methods;
- moved all implementations to "impl" package.

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
@@ -1,0 +1,209 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+
+    @Override
+    public Car create(Car car) {
+        String createRequest = "INSERT INTO cars(model, manufacturer_id)"
+                   + " values(?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createCarStatement = connection
+                        .prepareStatement(createRequest, Statement.RETURN_GENERATED_KEYS)) {
+            createCarStatement.setString(1, car.getModel());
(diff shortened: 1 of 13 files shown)


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String getCarByDriverRequest = "SELECT car_id FROM car_drivers WHERE driver_id = ?";
+        List<Car> cars = new ArrayList<>();
+        List<Long> carsId = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getCarsByDriverStatement = connection
+                        .prepareStatement(getCarByDriverRequest, Statement.RETURN_GENERATED_KEYS)) {
```
RomanDubovskyi, line 97: same here

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            getCarsByDriverStatement.setObject(1, driverId);
+            ResultSet resultSet = getCarsByDriverStatement.executeQuery();
+            while (resultSet.next()) {
+                Long carId = resultSet.getObject("car_id", Long.class);
+                carsId.add(carId);
```
RomanDubovskyi, line 102: we can use the same logic as in getAll method. Let's init cars without drivers here, and after try catch in forEach init drivers for each car

RomanDubovskyi (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    public Optional<Car> get(Long id) {
+        String getRequest = "SELECT cars.id, model, manufacturer_id, "
+                          + "manufacturers.id, name, country "
+                          + "FROM cars INNER JOIN manufacturers "
+                          + "ON manufacturers.id = cars.manufacturer_id "
+                          + "WHERE cars.id = ? AND cars.is_deleted = false;";
+        Car car = null;
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getCarStatement = connection
+                        .prepareStatement(getRequest, Statement.RETURN_GENERATED_KEYS)) {
```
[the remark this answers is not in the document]
ASelivonchyk: Ok, I understood mistake.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
ASelivonchyk: Why it will be better? In any way here will be two different select request to get car fields and driver id.

RomanDubovskyi (approved):

PR #187, opened 2022-01-31
Hw jv join main

maxkmaxk:


src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers() == null) {
+            car.setDrivers(new ArrayList<>());
+        }
```
MaliukDaria, line 47: I dont think we need this

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        car.getDrivers().add(driver);
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers() == null) {
+            throw new RuntimeException(car + " doesn't contain " + driver);
+        }
```
MaliukDaria, line 61: I dont think we need this

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #193, opened 2022-02-15
join example (sol) f

LinkClink:


MaliukDaria (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+            while (resultSet.next()) {
+                cars.add(getCarWithSpecificDriverId(resultSet));
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from driversDB.", e);
+        }
+    }
+
+    private Car getCarWithSpecificDriverId(ResultSet resultSet) throws SQLException {
```
MaliukDaria, line 83: private methods should be placed after public

MaliukDaria (approved):

PR #196, opened 2022-02-20
Implement jv join

pikador10:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
MaliukDaria, line 9:
```suggestion
    Optional<Car> get(Long id);
```

It's better to return Optional from dao layer in places, where we could get null

MaliukDaria (changes requested):

PR #199, opened 2022-03-03
Car: model, dao, service and DB table was implemented and tested successfully.

karma-o:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Inject
+    private ManufacturerService manufacturerService;
+
+    @Override
+    public Car create(Car car) {
+        String createCarQuery = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?)";
+        String createDriversRelationQuery =
+                "INSERT INTO cars_drivers(driver_id, car_id) VALUES (?, ?);";
```
MaliukDaria, line 29: I think it's better to move car-drivers relation creation in separate private method. If in the future we have new relationships for car and we need to add them to the create method, then this method will come out too large. And we could reuse it in update method. (same in update method)

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            for (Driver driver : car.getDrivers()) {
+                createDriversRelationStatement.setLong(1, driver.getId());
+                createDriversRelationStatement.setLong(2,car.getId());
+                createDriversRelationStatement.execute();
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create "
```
MaliukDaria, line 51:
```suggestion
            throw new DataProcessingException("Couldn't create car: "
```

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException("Could not get car "
+                        + "from DAO by id = " + id)
+                );
```
MaliukDaria, line 29:
```suggestion
                .orElseThrow(() -> new NoSuchElementException("There is no car with id = " + id)
                );
```

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    public void removeDriverFromCar(Driver driver, Car car) {
+        car.getDrivers().remove(driver);
+        carDao.update(car);
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        return carDao.getAll().stream()
+                .filter(car -> car.getDrivers().contains(driverService.get(driverId)))
+                .collect(Collectors.toList());
```
MaliukDaria, line 63: It's bad practice to get all cars from db and then filter them. What if you have 1 000 000 cars in db, then you have to get them all every time when you need to get drivers cars. Lets create separate method List<Car> getByDriverId(Long driverId); on dao layer for this purpose

MaliukDaria (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,237 @@
...
+        ) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(parseCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from cars table.", e);
+        }
+        for (Car c : cars) {
+            c.setDrivers(getDriversOfCar(c.getId()));
```
MaliukDaria, line 100:
```suggestion
        for (Car car : cars) {
            c.setDrivers(getDriversOfCar(c.getId()));
```

MaliukDaria (approved):

PR #202, opened 2022-03-10
Implementhed Car, CarDao, CarService

DaniilOpryshko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,49 @@
...
+        driverTwo = driverService.create(driverTwo);
+        List<Driver> drivers = new ArrayList<>();
+        drivers.add(driverOne);
+        Car bmwCar = new Car();
+        bmwCar.setModel("x6");
+        bmwCar.setManufacturer(bmw);
+        bmwCar.setDrivers(drivers);
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        bmwCar = carService.create(bmwCar);
+        System.out.println(carService.get(1L) + "\n");
```
MaliukDaria, line 42:
```suggestion
        System.out.println(carService.get(bmwCar.getId()) + "\n");
```

It's better not to use magic numbers, we can get Id from bmwCar entity

MaliukDaria (approved):

PR #219, opened 2022-03-24
was implemented methods car dao and car service

IgorZima90:


DenTima (approved):
I think everything is fine

boroda4436 (approved):

PR #234, opened 2022-05-11
added CarService class and CarDaoImpl class with all methods

Alina-Vasylyshyn:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        } catch (SQLException throwable) {
+            throw new RuntimeException("Can't delete driver from car: ", throwable);
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String selectRequest = "SELECT c.id AS car_id, model, m.id AS manufacturer_id, "
+                + "m.name, m.country FROM cars c JOIN manufacturers m "
+                + "ON c.manufacturer_id = m.id WHERE c.id = ? AND c.is_deleted = FALSE";
```
KostyaNikiforov, line 71:
```suggestion
        String selectRequest = "SELECT c.id AS car_id, model, m.id AS manufacturer_id, "
                + "m.name, m.country FROM cars c "
                + "JOIN manufacturers m ON c.manufacturer_id = m.id "
                + "WHERE c.id = ? AND c.is_deleted = FALSE";
```
It is better to move 'join' to the new line

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,50 @@
...
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        ManufacturerService manufacturerService = (ManufacturerService) injector
+                .getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = manufacturerService.get(2L);
```
KostyaNikiforov, line 20: What if someone else runs your code, will they get the same result?

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,50 @@
...
+        Driver driver1 = new Driver("Leo", "711199");
+        Driver driver2 = new Driver("Alexa", "922278");
```
KostyaNikiforov, line 22: It's not a best practice to name your variables like this

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Alina-Vasylyshyn: Ok

src/main/java/mate/jdbc/Main.java
Alina-Vasylyshyn: Yes I understood. No, there will be no similar result. I have to add first manufacturer,  to get the result correctly. In my table I have result for this methods, but someone else dont have (if dont use my  before task, when i add manufacture for this id).

src/main/java/mate/jdbc/Main.java
Alina-Vasylyshyn: Yes, I understood. Because later I forget who is driver1, who is driver2

KostyaNikiforov (approved):

PR #266, opened 2022-05-17
did HW 04

oleksandr-protsenko:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                            PreparedStatement createCarStatement =
+                          connection.prepareStatement(query, Statement.RETURN_GENERATED_KEYS)) {
+            createCarStatement.setLong(1, car.getManufacturer().getId());
```
Syon2154, line 26: Are you sure we need to set Id here?

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+            createCarStatement.setString(2, car.getModel());
+            createCarStatement.executeUpdate();
+            ResultSet generatedKeys = createCarStatement.getGeneratedKeys();
```
Syon2154, line 29: It would be better to replace `ResultSet generatedKeys` with `ResultSet generatedKey`. Because we expect to get only one key for new car

src/main/java/mate/jdbc/dao/CarDaoImpl.java
oleksandr-protsenko: yes?

Syon2154 (approved):

PR #277, opened 2022-06-19
jv-join was added

alexpogrebniy:


Nick97-git (changes requested):

Nick97-git (changes requested):
you have one unresolved comment from my previous review

alexpogrebniy:
I don't understand what you're talking about

alexpogrebniy:
removed the unused constructor

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Objects;
+
+public class Car {
+    private Long id;
+    private String model;
+    private Manufacturer manufacturer;
+    private List<Driver> drivers;
+
+    public Car() {
+    }
```
Nick97-git, line 13: if you removed the constructor with parameters, do you need to add the constructor without them explicitly?

Nick97-git (changes requested):

src/main/java/mate/jdbc/model/Car.java
alexpogrebniy: No, no need

Nick97-git (approved):

PR #280, opened 2022-06-22
Develop

Sm0kk:
Fixed after review


src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers().remove(driver)) {
+            carDao.update(car);
+        } else {
+            throw new RuntimeException("Couldn't remove driver " + driver + " from car " + car);
+        }
+    }
```
liliia-ponomarenko, line 61: What happens if a driver can't be removed?

liliia-ponomarenko (changes requested):

KostyaNikiforov (approved):

PR #298, opened 2022-07-06
Car, CarDao, CarService was impled and tested

oleksii-kordysh:


MKaramavrov (approved):
Good job 🔥

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -19,14 +19,10 @@ public class ConnectionUtil {
         }
     }
 
-    public static Connection getConnection() {
+    public static Connection getConnection() throws SQLException {
```
dSkarzh, line 22: better leave as it was, so we can have information specific to bad connection, not just general can't do something

src/main/java/mate/jdbc/TestServices.java
```diff
@@ -0,0 +1,57 @@
...
+    public static void testCarServices() {
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        //getAll() check
+        carService.getAll().forEach(System.out::println);
+
+        //getAllByDriver() check
+        carService.getAllByDriver(1L).forEach(System.out::println);
+
+        //getById() check
+        carService.get(7L);
```
dSkarzh, line 22: try not to use hardcoded ids

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,8 @@
 package mate.jdbc;
 
 public class Main {
+
     public static void main(String[] args) {
-        // test your code here
+        TestServices.testCarServices();
```
dSkarzh, line 6: you could test directly in main, it's ok

dSkarzh (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,8 +1,62 @@
...
+import mate.jdbc.model.Car;
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final long TEST_CAR_ID = 7L;
+    private static final long TEST_DRIVER_ID = 1L;
+    private static final long TEST_MANUFACTURER_ID = 1L;
+    private static final String TEST_MODEL = "test";
```
dSkarzh, line 14: that's still hardcoded, but okay
i meant something like `carService.delete(testCar.getId())`

dSkarzh (approved):

PR #299, opened 2022-07-06
CarDao and CarService were implemented

mykola8426:


NikitaSemisoshenko (approved):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+        return Optional.ofNullable(car);
+    }
+
+    private Car parseCarFromDb(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        car.setId(resultSet.getObject("id", Long.class));
+        car.setModel(resultSet.getString("model"));
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setId(resultSet.getLong("manufacturer_id"));
```
dSkarzh, line 85: better to use `getObject()`

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,22 @@ CREATE TABLE `drivers` (
...
+                        `manufacturer_id` bigint DEFAULT NULL,
+                        PRIMARY KEY (`id`),
+                        KEY `cars_manufacturers_fk` (`manufacturer_id`),
+                        CONSTRAINT `cars_manufacturers_fk` FOREIGN KEY (`manufacturer_id`) REFERENCES `manufacturers` (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=7 DEFAULT CHARSET=utf8mb3;
+
+CREATE TABLE `cars_drivers` (
+                                `driver_id` bigint NOT NULL,
+                                `car_id` bigint NOT NULL,
+                                `is_deleted` tinyint NOT NULL DEFAULT '0',
```
dSkarzh, line 34: remove this column, cars_drivers table is not an entity table, so we can just delete records from it, without soft delete

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+                + "ON cars.manufacturer_id = manufacturers.id "
+                + "WHERE cars.is_deleted = false;";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                Car car = parseCarFromDb(resultSet);
+                List<Driver> drivers = getDriversFromDb(resultSet.getLong("id"));
+                cars.add(car);
```
dSkarzh, line 137: only fetching drivers data from db is not enough

dSkarzh: you need to set drivers into car

dSkarzh: and you have nested connection problem here

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+                + "WHERE car_id = ? AND is_deleted = false;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setString(1, car.getModel());
+            statement.setLong(2, car.getManufacturer().getId());
+            statement.setLong(3, car.getId());
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't update car in db. Car: " + car, e);
+        } finally {
```
dSkarzh, line 158: finally is not needed here
just do it without it

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE cars SET is_deleted = false WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            int executeUpdate = statement.executeUpdate();
```
dSkarzh, line 185:
```suggestion
            int rowsAffected = statement.executeUpdate();
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        } finally {
+            deleteAllRelations(car);
+            insertDrivers(car);
+            return car;
+        }
+
+    }
+
+    private void deleteAllRelations(Car car) {
+        String query = "UPDATE cars_drivers "
```
dSkarzh, line 167: use `DELETE` here, there is no need to use soft delete strategy in cars_drivers

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, driverId);
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                Car car = new Car();
+                car.setId(resultSet.getLong("id"));
+                car.setModel(resultSet.getString("model"));
+                cars.add(car);
```
dSkarzh, line 208: think how you can refactor your code so we have only one private helper method that parses car (hint: let's fetch from db not only car, cars_drivers relations, but manufacturer as well)

dSkarzh (changes requested):
Some improvements needed

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+                Car car = new Car();
+                car.setId(resultSet.getLong("id"));
+                car.setModel(resultSet.getString("model"));
+                Manufacturer manufacturer = new Manufacturer();
+                manufacturer.setName(resultSet.getString("name"));
+                manufacturer.setCountry(resultSet.getString("country"));
+                manufacturer.setId(resultSet.getObject("manufacturer_id",
+                        Long.class));
+                car.setManufacturer(manufacturer);
+                cars.add(car);
```
dSkarzh, line 212: you have method that does this `parseCarFromDb()`
also move private methods to the bottom of the class

dSkarzh (changes requested):

PR #303, opened 2022-07-07
created car layer

DimaPodakin:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,199 @@
...
+                car = getCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Cannot find car by id: " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriverList(id));
+            return Optional.of(car);
+        }
+        return Optional.empty();
```
YaroslavLa, line 61: You can use ofNullable()

YaroslavLa (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
dSkarzh, line 61:
```suggestion
        if (car != null) {
            car.setDrivers(getDriverList(id));
        }
        return Optional.ofNullable(car);
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,199 @@
...
+                car.setDrivers(getDriverList(car.getId()));
+                carList.add(car);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Cannot get all cars by driver id: " + driverId, e);
+        }
+        return carList;
+    }
+
+    private void insertDriver(Car car) {
```
dSkarzh, line 138:
```suggestion
    private void insertDrivers(Car car) {
```

dSkarzh (changes requested):

dSkarzh (approved):

PR #307, opened 2022-07-12
JDBC join homework implemented

Illinois357:


src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).orElseThrow(() -> new NoSuchElementException("Could not get car "
```
MKaramavrov, line 23:
```suggestion
        return carDao.get(id).orElseThrow(() -> new DataProcessingException("Could not get car "
```
You could throw your custom exception here

MKaramavrov (changes requested):
Great solution! Just some minor fixes.

src/main/java/mate/jdbc/service/CarServiceImpl.java
Illinois357: I used the template from other implemented services given

MKaramavrov (approved):

KostyaNikiforov (approved):

PR #311, opened 2022-07-14
Hw jv join

Oleksii-Mospan:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,106 @@
...
     public static void main(String[] args) {
         // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+
+        List<Driver> driverList2 = new ArrayList<>();
+        List<Driver> driverList4 = new ArrayList<>();
+
+        driverList2.add(driverService.get(3L));
+        driverList4.add(driverService.get(2L));
+        driverList4.add(driverService.get(4L));
```
MaliukDaria, line 25: why we could get driver from db if we have not add it before?

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
MaliukDaria, line 9:
```suggestion
    Optional<Car> get(Long id);
```

we should return Optional for `get(Long id)` method in Dao layer

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+            statement.setLong(2, car.getManufacturer().getId());
+            statement.setLong(3, car.getId());
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update " + car + " in carsDB.", e);
+        }
+        deleteAllRelationsForCarQuery(car.getId());
+        if (car.getDrivers() != null) {
+            insertRelationForCarQuery(car.getId(), car.getDrivers());
+        }
```
MaliukDaria, line 98:
```suggestion
        if (car.getDrivers() != null) {
            deleteAllRelationsForCarQuery(car.getId());
            insertRelationForCarQuery(car.getId(), car.getDrivers());
        }
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #317, opened 2022-07-29
created class Car and CarServise interface with implementation

vova5179:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+        String query = "INSERT INTO cars_drivers (car_id, driver_id) VALUES(?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                statement.setLong(2, driver.getId());
+                statement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't insert drivers to car", e);
```
MaliukDaria, line 216:
```suggestion
            throw new DataProcessingException("Couldn't insert drivers to car " + car, e);
```

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException("Could not get car "
+                + "by id = " + id));
```
MaliukDaria, line 25:
```suggestion
        return carDao.get(id)
                .orElseThrow(() -> new NoSuchElementException("There is no car "
                + "with id = " + id));
```

MaliukDaria (approved):

PR #340, opened 2022-09-11
Jv join

ira0807:
Create class Car
Create CarService interface with implementation
Create CarServiceDao interface with implementation

### DIFF
--- a/src/main/java/mate/jdbc/dao/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/CarDaoImpl.java
@@ -1,0 +1,224 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String insertQuery = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createCarStatement = connection.prepareStatement(insertQuery,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            createCarStatement.setLong(1, car.getManufacturer().getId());
+            createCarStatement.setString(2, car.getModel());
+            createCarStatement.executeUpdate();
+            ResultSet generatedKeys = createCarStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1, Long.class);
+                car.setId(id);
+            }
+        } catch (SQLException e) {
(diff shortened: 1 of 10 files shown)


max-bolotin (approved):
good job!

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String insertQuery = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createCarStatement = connection.prepareStatement(insertQuery,
+                        Statement.RETURN_GENERATED_KEYS)) {
```
dSkarzh, line 24:
```suggestion
        String query = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?);";
        try (Connection connection = ConnectionUtil.getConnection();
                PreparedStatement statement = connection.prepareStatement(insertQuery,
                        Statement.RETURN_GENERATED_KEYS)) {
```
use simpler naming, it is obvious from the context what they are for

dSkarzh: this goes for all queries and statement, not only here

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+            updateStatement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update car: " + car, e);
+        }
+        deleteCarsWithDrivers(car.getId());
+        insertCarsWithDrivers(car);
+        return car;
+    }
+
+    private void insertCarsWithDrivers(Car car) {
```
dSkarzh, line 104:
```suggestion
    private void insertDriversForCar(Car car) {
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+            for (Driver driver : car.getDrivers()) {
+                insertStatement.setLong(2, driver.getId());
+                insertStatement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't insert drivers into car " + car, e);
+        }
+    }
+
+    private int deleteCarsWithDrivers(Long id) {
```
dSkarzh, line 119:
```suggestion
    private int deleteDriversFromCar(Long id) {
```

dSkarzh: also, let's move these private methods to the end of the class

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private Manufacturer parseManufacturer(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setId(resultSet.getObject("manufacturer_id", Long.class));
+        manufacturer.setName(resultSet.getString("name"));
+        manufacturer.setCountry(resultSet.getString("country"));
+        return manufacturer;
+    }
+
+    private void insertDrivers(Car car) {
```
dSkarzh, line 210: you already have method that does this, why do you need a second one?

dSkarzh (changes requested):

dSkarzh (approved):

PR #357, opened 2022-09-12
created class Car, implemented CarDao, CarDaoImpl, CarService, CarSer…

ArturZhadan:
…viceImpl

### DIFF
--- a/src/main/java/mate/jdbc/dao/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/CarDaoImpl.java
@@ -1,0 +1,226 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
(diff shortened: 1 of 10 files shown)


Toltbcm (approved):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String insertRequest = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?)";
```
dSkarzh, line 21:
```suggestion
        String query = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?)";
```
it is obvious from the context what this query does, so you can simply call it query

dSkarzh: same goes for all other queries

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                PreparedStatement statement = connection.prepareStatement(insertDriversRequest)) {
+            statement.setLong(1, car.getId());
+            for (Driver driver: car.getDrivers()) {
+                statement.setLong(2, driver.getId());
+                statement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t insert drivers to car", e);
+        }
+    }
```
dSkarzh, line 155: why do you need this, when you have `insertCarDrivers()`

dSkarzh (changes requested):

dSkarzh (approved):

PR #366, opened 2022-09-13
create tables cars and cars_drivers, create and implements classes Dr…

veers-droid:
…iver, DriverDaoImpl, DriverService, write some tests in Main

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
@@ -1,0 +1,210 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?);";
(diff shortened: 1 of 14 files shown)


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id);
```
justpopova, line 23:
```suggestion
        return carDao.get(id).orElseThrow(
                () -> new NoSuchElementException("Can't get car by this id: " + id));
```
cause u return optional from dao, so need to throw exception in case if there no car

justpopova (changes requested):
Good Job!

justpopova (approved):
🔥

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            throw new DataProcessingException("Couldn't update car: " + car, e);
+        }
+        deleteDriversFromCar(car.getId());
+        insertDriversForCar(car);
+        return car;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE cars SET is_deleted = TRUE where id = ?";
```
dSkarzh, line 107:
```suggestion
        String query = "UPDATE cars SET is_deleted = TRUE WHERE id = ?";
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            return statement.executeUpdate() != 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT c.id as car_id, model, manufacturer_id, name, "
+                + "country FROM cars_drivers cd JOIN cars c on c.id = cd.car_id "
```
dSkarzh, line 121: would make more sense to make `cars` root table, not `cars_drivers`

dSkarzh (approved):

PR #372, opened 2022-09-13
implemented car service functionality

AshaJaneBrown:


Aptypio13 (approved):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,198 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                         PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, driverId);
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+            for (Car car : cars) {
+                car.setDrivers(getDriversForCar(car));
+            }
```
dSkarzh, line 128: nested connection problem here

dSkarzh (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
AshaJaneBrown: @dSkarzh 
Do I need to create a separate method for that purpose?

dSkarzh (approved):

PR #380, opened 2022-09-14
Car's Dao and Service was implemented

not4und:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,182 @@
...
+                cars.add(getCar(resultSet, connection));
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from carsDB "
+                    + "by driver id " + driverId, e);
+        }
+    }
+
+    private Car insertDrivers(Car car, Connection connection) throws SQLException {
```
zachar-chorny, line 129: No need to pass it on here, open a new one here

zachar-chorny (changes requested):

Rommelua (approved):

PR #394, opened 2022-10-19
jv-join done

Aalderaan:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,53 @@
...
+
 public class Main {
+
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturerToyota = new Manufacturer(null, "Toyota", "Japan");
```
MaliukDaria, line 20:
```suggestion
        Manufacturer manufacturerToyota = new Manufacturer("Toyota", "Japan");
```

It is bad practice to pass null to constructor

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String getQuery = "SELECT * FROM cars c JOIN manufacturers m"
+                + " ON c.manufacturer_id = m.id"
+                + " WHERE c.id = ? AND c.is_deleted = FALSE AND m.is_deleted = FALSE;";
```
MaliukDaria, line 45: Why we cant show car if we delete manufacturer? Do we delete car when delete manufacturer?

MaliukDaria (approved):
The task should not be implemented in the master/main branch, a separate branch should be created for this (you can leave this task as it is, but create separate branches for next onews 🙂)

src/main/java/mate/jdbc/Main.java
Aalderaan: I did it on master by mistake) Will be accurate next time. Thank You Daria. Ok, will change it!

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Aalderaan: Yes, you are right) It must be possible to see the car if manufacturer is deleted. Will change it as well

PR #401, opened 2022-11-02
jv-join

SoulARC:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,205 @@
...
+                + "WHERE c.id = ? AND c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            ResultSet resultSet = statement.executeQuery();
+            Car car = null;
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+            return Optional.ofNullable(car);
```
EvgenyTyshchenko, line 54: Probabiy we should set car drivers here
        if (car != null) {
            car.setDrivers(getCarDrivers(id));
        }
I'm not sure, so I suggest leaving this question for mentors 
review

EvgenyTyshchenko (approved):
LGTM

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,205 @@
...
+            throw new DataProcessingException("Couldn't delete drivers from car: " + car, e);
+        }
+    }
+
+    private void addDriversToCar(List<Driver> list, Long carId) {
+        String query = "INSERT INTO cars_drivers (driver_id, car_id) VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            for (Driver driver : list) {
```
KostyaNikiforov, line 195:
```suggestion
            for (Driver driver : drivers) {
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
KostyaNikiforov, line 54:
```suggestion
            if (resultSet.next()) {
                return Optional.of(getCar(resultSet));
            }
            return Optional.empty();
```

KostyaNikiforov (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
SoulARC: It already happened, I forgot. Inattentive(

PR #410, opened 2022-11-08
HW 4 v.1.0

ghost1975:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+    public List<Car> getAllByDriver(Long id) {
+        List<Car> cars = new ArrayList<>();
+        String getAllByDriverQuery =
+                "SELECT joined_table.car_id, joined_table.model, manufacturer_id, "
+                        + "manufacturers.name, manufacturers.country "
+                        + "FROM manufacturers "
+                        + "INNER JOIN ("
+                        + "SELECT * FROM cars "
+                        + "INNER JOIN drivers_cars ON cars.id = drivers_cars.car_id"
+                        + ") as joined_table ON manufacturers.id = joined_table.car_id;";
```
Rommelua, line 132: better use something like this. SELECT inside JOIN is not a good practice.
```
"FROM cars c"
                + " JOIN manufacturers m ON c.manufacturer_id = m.id"
                + " JOIN cars_drivers cd ON c.id = cd.car_id"
                + " JOIN drivers d ON cd.driver_id = d.id"
                + " WHERE c.is_deleted = FALSE AND driver_id = ?"
                + " AND d.is_deleted = FALSE";
```

Rommelua (changes requested):

PR #411, opened 2022-11-09
Realized join connection between car and driver also drivers-cars man…

anastasiavasylyk:
…y-to-many connection

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,12 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.CarService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService = (CarService) injector.getInstance(CarService.class);
     }
 }
--- a/src/main/java/mate/jdbc/dao/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/CarDaoImpl.java
@@ -1,0 +1,213 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createCarStatement = connection.prepareStatement(query,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            createCarStatement.setLong(1, car.getManufacturer().getId());
+            createCarStatement.setString(2, car.getModel());
+            createCarStatement.executeUpdate();
+            ResultSet generatedKeys = createCarStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                car.setId(generatedKeys.getObject(1, Long.class));
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create "
+                    + car + ". ", e);
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT c.id AS car_id, manufacturers.id AS manufacturer_id,"
+                + " manufacturers.name AS manufacturer_name, manufacturers.country"
+                + " AS manufacturer_country, model FROM cars c"
+                + " JOIN manufacturers  ON c.manufacturer_id = manufacturers.id"
+                + " WHERE c.id = ? AND c.is_deleted = FALSE;";
+        Car car = null;
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getCarStatement = connection.prepareStatement(query,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            getCarStatement.setLong(1, id);
+            ResultSet resultSet = getCarStatement.executeQuery();
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+            assert car != null;
+            car.setDrivers(getCarDrivers(id));
+            return Optional.of(car);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car by id " + id + ". ", e);
+        }
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT c.id AS car_id, manufacturers.id AS manufacturer_id,"
+                + " manufacturers.name AS manufacturer_name,"
+                + " manufacturers.country AS manufacturer_country, model "
+                + "FROM cars c JOIN manufacturers  ON c.manufacturer_id = manufacturers.id "
+                + "WHERE c.is_deleted = FALSE;";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllCarsStatement = connection.prepareStatement(query)) {
+            ResultSet resultSet = getAllCarsStatement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from carsDB.", e);
+        }
+    }
+
+    @Override
+    public Car update(Car car) {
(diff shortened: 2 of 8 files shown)


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getCarStatement = connection.prepareStatement(query,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            getCarStatement.setLong(1, id);
+            ResultSet resultSet = getCarStatement.executeQuery();
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+            assert car != null;
+            car.setDrivers(getCarDrivers(id));
```
shmex-j, line 57: I think it's better to call getCarDrivers() after try-catch, because this method opens new connection, and as @boroda4436 said in topic, it's better to close previous connection before opening another one

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllCarsStatement = connection.prepareStatement(query)) {
+            ResultSet resultSet = getAllCarsStatement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from carsDB.", e);
+        }
```
shmex-j, line 81: There's you didn't included car drivers in cars. I think you should add drivers to cars in foreach after cars list is filled and connection is closed

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+            updateCarStatement.setLong(1, car.getManufacturer().getId());
+            updateCarStatement.setString(2, car.getModel());
+            updateCarStatement.setLong(3, car.getId());
+            updateCarStatement.executeUpdate();
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update "
+                    + car + " in carsDB.", e);
+        }
+    }
```
shmex-j, line 101: There, when you update car, you should remove all drivers from it, and then add them all again to update

src/main/java/mate/jdbc/dao/CarDaoImpl.java
shmex-j: I see you calling getCarDrivers in getCar method, which is not good because you create new connection to db while old one still opened

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,12 @@
...
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.CarService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService = (CarService) injector.getInstance(CarService.class);
```
shmex-j, line 10: Maybe you should test your solution here as in DaoService and JDBC task

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+                PreparedStatement removeDriverFromCarStatement
+                        = connection.prepareStatement(query)) {
+            removeDriverFromCarStatement.setLong(1, driver.getId());
+            removeDriverFromCarStatement.setLong(2, car.getId());
+            removeDriverFromCarStatement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete driver from car with id "
+                    + driver.getId(), e);
+        }
+    }
```
shmex-j, line 142: This methods [should exist only on service layer, not in dao](https://mate-academy.github.io/jv-program-common-mistakes/java-JDBC/join/Joins_checklist.html)

shmex-j (changes requested):
Some suggestions

src/main/java/mate/jdbc/Main.java
anastasiavasylyk: I have tested my methods and remove after that, cause mentor in previous task tell me to delete
Знімок екрана 2022-11-09 120909

shmex-j (approved):
LGTM🔥

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                    + car + " in carsDB.", e);
+        }
+        removeDriversFromCar(car);
+        addCarDrivers(car.getDrivers(), car);
+        return car;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE cars SET is_deleted = TRUE WHERE id = ?";
```
liliia-ponomarenko, line 108:
```suggestion
        String query = "UPDATE cars SET is_deleted = TRUE WHERE id = ? AND is_deleted = FALSE";
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT c.id AS car_id, manufacturer_id, model "
+                + "FROM cars c JOIN cars_drivers cd ON c.id = cd.car_id"
+                + " WHERE cd.driver_id = ? AND c.is_deleted = FALSE;";
```
liliia-ponomarenko, line 122: You also should check if the driver is not deleted

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            ResultSet resultSet = getAllCarsByDriverStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                Car car = new Car();
+                car.setId(resultSet.getObject("car_id", Long.class));
+                Manufacturer manufacturer = new Manufacturer();
+                manufacturer.setId(resultSet.getObject("manufacturer_id", Long.class));
+                car.setManufacturer(manufacturer);
+                car.setModel(resultSet.getString("model"));
+                cars.add(car);
```
liliia-ponomarenko, line 136: better move this logic to the private method

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get cars by driver id " + driverId, e);
+        }
```
liliia-ponomarenko, line 141: You also should  add drivers to all cars

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement addCarDriversStatement
+                        = connection.prepareStatement(insertCarDriverSql)) {
+            for (Driver driver : drivers) {
+                addCarDriversStatement.setLong(1, driver.getId());
+                addCarDriversStatement.setLong(2, car.getId());
+                addCarDriversStatement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
```
liliia-ponomarenko, line 212: Don’t leave exceptions without informative messages

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public Car(String model, Manufacturer manufacturer) {
+        this.model = model;
+        this.manufacturer = manufacturer;
+    }
+
+    public Car(Long id, String model, Manufacturer manufacturer, List<Driver> drivers) {
+        this.id = id;
+        this.model = model;
```
liliia-ponomarenko, line 29: Do you really need all these constructors?

liliia-ponomarenko (changes requested):
Good job! Also, you should supplement ```init_db```

src/main/java/mate/jdbc/Main.java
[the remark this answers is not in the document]
anastasiavasylyk: я це зрозуміла ще мунулого завдання, і я тестувала весь функціонал, тому впевнена, що все працює, щодо того, що тут зараз немає ніякого тестування, то, думаю це нічого сташного, оскільки все протестоване і суть завдання було написати цей функціонал

src/main/java/mate/jdbc/dao/CarDaoImpl.java
anastasiavasylyk: there is check ...AND c.is_deleted = FALSE;

liliia-ponomarenko: yes, you check that car is not deleted but the driver  also can be deleted 
so you need something  like this AND d.is_deleted = FALSE (you will also need to join a driver table)

liliia-ponomarenko (changes requested):
Answered your question 🙋‍♀️

KostyaNikiforov (approved):

PR #423, opened 2022-11-15
created solution for join task

Bohdan20b:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,192 @@
...
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+
```
GrishaSeniv, line 20: empty line

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,192 @@
...
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create car: "
+                    + car, e);
+        }
```
GrishaSeniv, line 38: You need to insert drivers for car too

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,192 @@
...
+    public Optional<Car> get(Long id) {
+        String query = "SELECT c.id, manufacturer_id, model, name, country "
+                + "FROM cars c JOIN manufacturers m ON c.manufacturer_id = m.id "
+                + "WHERE c.id = ? AND c.is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            ResultSet resultSet = statement.executeQuery();
+            if (resultSet.next()) {
+                return Optional.of(getCar(resultSet));
```
GrishaSeniv, line 51: You open the connection twice, maybe it would be better to get car from first query, and after this  try-catch, u can use method to set drivers to car. But i'm not sure, it's optional
P.S Bohdan told about this in videos

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,14 @@ CREATE TABLE `drivers` (
...
+                        `id` bigint NOT NULL AUTO_INCREMENT,
+                        `manufacturer_id` int DEFAULT NULL,
+                        `model` varchar(255) DEFAULT NULL,
+                        `is_deleted` tinyint DEFAULT 0,
+                        PRIMARY KEY (`id`)
+);
+CREATE TABLE `cars_drivers` (
+                                `driver_id` bigint DEFAULT NULL,
+                                `car_id` int DEFAULT NULL
+);
```
GrishaSeniv, line 31: Don't forget to add an empty line at the end of each file

GrishaSeniv (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Bohdan20b: i add them in the addDriverToCar() method via CarService, it also works alright.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Bohdan20b: Let's leave it for the mentors, because i'm also not sure

KostyaNikiforov (approved):
🔥

PR #428, opened 2022-11-15
Implemented: CarService, CarDao classes

NSay0:


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).orElseThrow(
+                () -> new RuntimeException("There is no such car with id:" + id));
```
yehbreus, line 24: i think that will be better throw NoSuchElementException here

yehbreus (changes requested):
Nice solutions!)

yehbreus (approved):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -19,8 +20,10 @@ public class DriverDaoImpl implements DriverDao {
     public Driver create(Driver driver) {
         String query = "INSERT INTO drivers (name, license_number) VALUES (?, ?)";
         try (Connection connection = ConnectionUtil.getConnection();
-                PreparedStatement statement = connection.prepareStatement(query,
-                        Statement.RETURN_GENERATED_KEYS)) {
+                PreparedStatement statement = connection.prepareStatement(
+                        query,
+                        Statement.RETURN_GENERATED_KEYS
+                )) {
```
ChumakAn, line 26: it's better not to change style in already implemented part

ChumakAn (changes requested):
Looks good to me! Small comment

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query =
+                "SELECT c.id as car_id, c.model as car_model, mf.id as manufacturer_id, "
+                        + "mf.name as manufacturer_name, mf.country as manufacturer_country "
+                        + "FROM cars c JOIN manufacturers mf ON c.manufacturer_id = mf.id "
+                        + "WHERE c.id = ? AND c.is_deleted = FALSE;";
```
ChumakAn, line 47: also, it's better to make queries more readable - just format them

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
NSay0: mb was applied autoFormatting for some misstake.
+ Changed to previous look.

ChumakAn (approved):
It's better not to change the initial files, if it's not needed from the task

PR #437, opened 2022-11-16
Car dao and service implemented.

luka-lozovyk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,64 @@
...
+        CarService carService = (CarService)
+                injector.getInstance(CarService.class);
+        Car porsche911 = new Car("911", manufacturerPorsche, drivingPorscheList);
+        Car vwJetta = new Car("Jetta", manufacturerVW, drivingVwList);
+        carService.create(porsche911);
+        carService.create(vwJetta);
+
+        System.out.println("Car dao and service testing:");
+        System.out.println("Currently in db");
+        System.out.println(carService.getAll());
```
Iryna-a, line 47: Maybe it would be better to print all your cars like this: carService.getAll().forEach(System.out::println); This will improve readability

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,64 @@
...
+
+        vwJetta.setModel("Passat");
+        carService.update(vwJetta);
+        Car updated = carService.get(2L);
```
Iryna-a, line 51: It is better to use carService.get(carName.getId()) instead of hardcoded values because you can't be sure that the person who uses your code has the same information in DB tables as you(maybe the car which you updated has another id)

Iryna-a (changes requested):

Iryna-a (approved):
Good job!

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        return Optional.ofNullable(car);
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT c.id, c.model, c.manufacturer_id, mn.name, mn.country "
+                + "FROM cars c "
+                + "JOIN manufacturers mn ON c.manufacturer_id = mn.id "
+                + "WHERE c.is_deleted = FALSE;";
+        List<Car> allCars = new ArrayList<>();
```
ChumakAn, line 69:
```suggestion
        List<Car> cars = new ArrayList<>();
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, driverId);
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                allCars.add(getCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get all cars by riverId " + driverId, e);
+        }
+        for (Car car : allCars) {
```
ChumakAn, line 133:
```suggestion
        for (Car car : cars) {
```

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,70 @@
...
+        this.model = model;
+        this.manufacturer = manufacturer;
+        this.drivers = drivers;
+    }
+
+    public Car(Long id, String model, Manufacturer manufacturer, List<Driver> drivers) {
+        this.id = id;
+        this.model = model;
+        this.manufacturer = manufacturer;
+        this.drivers = drivers;
```
ChumakAn, line 28: we can use constructor overloading with this keyword

src/main/resources/init_db.sql
```diff
@@ -3,18 +3,43 @@ USE `taxi_service`;
...
-                                  `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                                  PRIMARY KEY (`id`),
-                                  UNIQUE INDEX `id_UNIQUE` (id ASC) VISIBLE,
-                                  UNIQUE INDEX `license_number_UNIQUE` (`license_number` ASC) VISIBLE);
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(225) NOT NULL,
+  `license_number` varchar(225) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `id_UNIQUE` (`id`),
```
ChumakAn, line 20: id is already unique so it's redundant

ChumakAn (changes requested):

ChumakAn (approved):

PR #440, opened 2022-11-17
hw-join

Iryna-a:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,225 @@
...
+            statement.setLong(1, id);
+            return statement.executeUpdate() != 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT car_id FROM cars_drivers WHERE driver_id = ?";
```
ChumakAn, line 121: maybe it's better to use joins on car_drivers and get a car from result set , using a method to parse it

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,92 @@
...
+        this.model = model;
+        this.manufacturer = manufacturer;
+    }
+
+    public Car(Long id, String model, Manufacturer manufacturer, List<Driver> drivers) {
+        this.id = id;
+        this.model = model;
+        this.manufacturer = manufacturer;
+        this.drivers = drivers;
+    }
```
ChumakAn, line 29: we can use constructor  overriding using `this`

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    public List<Car> getAllByDriver(Long driverId) {
+        return carDao.getAllByDriver(driverId);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = new ArrayList<>(car.getDrivers());
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
```
ChumakAn, line 53: we can simplify this

ChumakAn (changes requested):
Good! Minor improvements

ChumakAn (approved):

PR #446, opened 2022-11-17
added Car model and service, dao for working with this model and db

Danyl-Kniaziev:


src/main/resources/init_db.sql
```diff
@@ -18,3 +18,26 @@ CREATE TABLE `drivers` (
...
+
+CREATE TABLE `cars_drivers` (
+                                `car_id` BIGINT(11) NOT NULL,
+                                `driver_id` BIGINT(11) NOT NULL,
+                                CONSTRAINT `cars_dirvers_car_fk`
+                                    FOREIGN KEY (`car_id`)
+                                        REFERENCES `cars`(`id`),
+                                CONSTRAINT `cars_dirvers_driver_fk`
+                                    FOREIGN KEY (`driver_id`)
+                                        REFERENCES `drivers` (`id`));
```
romangolovatyi, line 43: we need added new line after code

romangolovatyi (approved):
Great!!

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -19,14 +19,10 @@ public class ConnectionUtil {
...
-    public static Connection getConnection() {
+    public static Connection getConnection() throws SQLException {
         Properties dbProperties = new Properties();
         dbProperties.setProperty("user", USERNAME);
         dbProperties.setProperty("password", PASSWORD);
-        try {
-            return DriverManager.getConnection(URL, dbProperties);
-        } catch (SQLException e) {
-            throw new RuntimeException("Can't create connection to DB ", e);
-        }
```
ChumakAn, line 30: it's better not to replace it with throws

ChumakAn: don't replace try-catch with throws

ChumakAn (changes requested):

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -26,7 +27,7 @@ public static Connection getConnection() {
         try {
             return DriverManager.getConnection(URL, dbProperties);
         } catch (SQLException e) {
-            throw new RuntimeException("Can't create connection to DB ", e);
+            throw new DataProcessingException("Can't create connection to DB ", e);
```
ChumakAn, line 30: it's not a Data Processing exception, because we will throw this exception in case the database connection cannot be established. Let's don't change this line and throw RuntimeException

ChumakAn (changes requested):

ChumakAn (approved):

PR #447, opened 2022-11-19
created class "Car" and "CarService"

Yaroslav781:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t create a car:" + car, e);
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT cars.id AS car_id, manufacturer_id, "
+                + "manufacturers.name AS manufacturer_name, "
+                + "manufacturers.country AS manufacturer_country, model\n"
```
GHSpeedWagon, line 43: we dont need "\n"

delete it in all cases

GHSpeedWagon (changes requested):
little fix

GHSpeedWagon (approved):
Gj

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -7,6 +7,14 @@ public class Driver {
     private String name;
     private String licenseNumber;
 
+    public Driver() {
+    }
+
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
hronograf, line 16: can we use `this(...)` here?

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -7,6 +7,14 @@ public class Manufacturer {
     private String name;
     private String country;
 
+    public Manufacturer() {
+    }
+
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
```
hronograf, line 16: same

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t create a car:" + car, e);
+        }
+    }
```
hronograf, line 37: what if we have some drivers for our car
should we save them to db?

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+            while (resultSet.next()) {
+                drivers.add(parseDrivers(resultSet));
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get drivers from car by id: " + carId, e);
+        }
+    }
+
+    private Driver parseDrivers(ResultSet resultSet) throws SQLException {
```
hronograf, line 165:
```suggestion
    private Driver parseDriver(ResultSet resultSet) throws SQLException {
```

hronograf (changes requested):
Good job!

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -1,13 +1,12 @@
...
 
 import java.util.List;
-import java.util.Optional;
 import mate.jdbc.model.Car;
 
 public interface CarDao {
     Car create(Car car);
 
-    Optional<Car> get(Long id);
+    Car get(Long id);
```
hronograf, line 9: no, here should be `Optional`, because car with such id may not exist, 
but on the service layer we should handle this situation and throw an exception or return another car or something else

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -30,14 +29,15 @@ public Car create(Car car) {
             if (resultSet.next()) {
                 car.setId(resultSet.getObject(1, Long.class));
             }
+            addDriverToCar(car);
```
hronograf, line 32: it would be better to open new connection after current connection will be closed, so lets move `addDriverToCar(car);` below try-catch block

src/main/resources/init_db.sql
```diff
@@ -23,7 +23,7 @@ CREATE TABLE `cars` (
                         `id` BIGINT NOT NULL AUTO_INCREMENT,
                         `model` VARCHAR(255) NOT NULL,
                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                        `manufacturer_id` carsINT NOT NULL,
+                        `manufacturer_id` INT NOT NULL,
```
hronograf, line 26: but `manufacturers.id` has type `BIGINT(11)`, 
so `manufacturer_id` also should be of type `BIGINT(11)`

hronograf (changes requested):

hronograf (approved):

PR #456, opened 2023-01-03
Added CarDao interface with implementation. Also created model Car. I…

AntonKalenskiy:
…n Service package created CarService interface with implementation.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,15 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.CarService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService =
+                (CarService) injector.getInstance(CarService.class);
+        carService.getAll().forEach(System.out::println);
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,15 @@
...
+import mate.jdbc.service.CarService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService =
+                (CarService) injector.getInstance(CarService.class);
+        carService.getAll().forEach(System.out::println);
```
zachar-chorny, line 12: Is getAll() only one method in car service? Lets test all methods here

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create car: "
+                    + car + ". ", e);
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    private void insertDrivers(Car car) {
```
zachar-chorny, line 43: all private methods should be after publics

zachar-chorny (changes requested):

aaananas (approved):

PR #463, opened 2023-01-12
added some test features for sql auto configuration

roman-yuskevych:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,195 @@
...
+                    + "from cars table.", e);
+        }
+        for (Car car : cars) {
+            car.setDrivers(getDriversForCar(car.getId()));
+        }
+        return cars;
+    }
+
+    @Override
+    public Car update(Car car) {
```
nacenik, line 94: Also, add logic to remove all drivers from the car, and add them from object, that you get into the method. Cause update means updating every single field in our class.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,195 @@
...
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a car by id " + id, e);
+        }
+    }
+
+    @Override
+    //TO DO
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT *";
```
nacenik, line 124: @here is the whole query with join?

src/main/resources/init_db.sql
```diff
@@ -3,18 +3,44 @@ USE `taxi_service`;
...
+        REFERENCES `taxi_service`.`manufacturers` (`id`)
+        ON DELETE NO ACTION ON UPDATE NO ACTION
+);
+
+DROP TABLE IF EXISTS `cars_drivers`;
+CREATE TABLE `cars_drivers` (
+    `car_id` BIGINT(11) NOT NULL,
+    `driver_id` BIGINT(11) NOT NULL,
+    CONSTRAINT `cars_drivers_cars_fk` FOREIGN KEY (`car_id`)
+        REFERENCES `taxi_service`.`cars` (`id`)
```
nacenik, line 41:
```suggestion
        REFERENCES `cars` (`id`)
```

src/main/resources/init_db.sql
```diff
@@ -3,18 +3,44 @@ USE `taxi_service`;
...
+        ON DELETE NO ACTION ON UPDATE NO ACTION,
+    CONSTRAINT `cars_drivers_drivers_fk` FOREIGN KEY (`driver_id`)
+        REFERENCES `taxi_service`.`drivers` (`id`)
```
nacenik, line 44:
```suggestion
        REFERENCES `drivers` (`id`)
```

nacenik (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,87 @@
...
+        //WARNING!!!
+
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(new Driver(null, "Abdul", "AA-479974"));
+        driverService.create(new Driver(null, "Hassan", "AH-272727"));
+        driverService.create(new Driver(null, "Mahmud", "AM-386921"));
+
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        manufacturerService.create(new Manufacturer(null, "Bentley", "England"));
```
okuzan, line 30: if you always pass null to id maybe create constructor without it

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,87 @@
...
+        manufacturerService.create(new Manufacturer(null, "Lexus", "Japan"));
+        manufacturerService.create(new Manufacturer(null, "Mercedes-Benz", "Germany"));
+
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        Car testCarFirst = new Car();
+        testCarFirst.setManufacturer(manufacturerService.get(3L));
```
okuzan, line 36: better save references to created manufactures and here call bentlyManufacturer.getId()

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,87 @@
...
+        testCarFirst.setModel("C 300");
+        carService.create(testCarFirst);
+        Car testCarSecond = new Car();
```
okuzan, line 39: bad naming, name it lincoln, lexus etc.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,87 @@
...
     public static void main(String[] args) {
-        // test your code here
+        System.out.println("Magic STARTS..." + System.lineSeparator());
+
+        //WARNING!!! Only for tests. This reset all schema to initial (empty tables) state.
+        ResetDbUtil.resetDbToInitialState();
+        //WARNING!!!
+
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(new Driver(null, "Abdul", "AA-479974"));
```
okuzan, line 24: if you always pass null to id maybe create constructor without it

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,87 @@
...
+        Car testCarSecond = new Car();
+        testCarSecond.setManufacturer(manufacturerService.get(1L));
+        testCarSecond.setModel("Continental GT Coupe V8");
+        carService.create(testCarSecond);
+        Car testCarThird = new Car();
+        testCarThird.setManufacturer(manufacturerService.get(2L));
+        testCarThird.setModel("Lexus LC Convertible");
+        carService.create(testCarThird);
+
+        carService.addDriverToCar(driverService.get(2L), carService.get(1L));
```
okuzan, line 48: relying heavily on magic numbers in main(), see if you can make them constants or retrieve from objects themselves

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                 PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, car.getManufacturer().getId());
+            statement.setString(2, car.getModel());
+            statement.setLong(3, car.getId());
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update a car " + car, e);
+        }
+        removeDriversFromCar(car);
+        addDriversToCar(car);
```
okuzan, line 107: I'd rather move this inside try block

okuzan (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,117 @@
...
+        System.out.println(System.lineSeparator()
+                + "Get random car from DB: " + System.lineSeparator()
+                + randomCarFromDb);
+
+        randomCarFromDb.getManufacturer().setName("Roman Yuskevych");
+        randomCarFromDb.getManufacturer().setCountry("Ukraine");
+        randomCarFromDb.setModel("BestModelEver");
+        manufacturerService.update(randomCarFromDb.getManufacturer());
+        carService.update(randomCarFromDb);
+        while (randomCarFromDb.getDrivers().size() > 0) {
```
aleksandr-bevza, line 95: Why do we use loop here for 1 iteration?
```suggestion
        if (randomCarFromDb.getDrivers().size() > 0) {
```

aleksandr-bevza (approved):

PR #479, opened 2023-01-25
task was completed, build success

Bohdan-Halitskyi:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create car of " + car, e);
+        }
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
```
Dmytro-Mk, line 42: once again, what about drivers? you return car with drives field = null when there could be drivers connected to car in db

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                car = getCar(resultSet);
+            }
+            return Optional.ofNullable(car);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get car by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAll() {
```
Dmytro-Mk, line 63: again cars with no drivers... 😞

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE cars SET is_deleted = TRUE "
+                + "WHERE id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete car by id " + id,e);
```
Dmytro-Mk, line 109:
```suggestion
            throw new DataProcessingException("Can't delete car by id " + id, e);
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT * FROM cars INNER JOIN manufacturers "
+                + "ON cars.manufacturer_id = manufacturers.id "
+                + "INNER JOIN cars_drivers ON cars.id = cars_drivers.car_id "
+                + "WHERE cars_drivers.driver_id = ?";
```
Dmytro-Mk, line 118:
```suggestion
        String query = "SELECT * FROM cars c JOIN manufacturers m "
                + "ON c.manufacturer_id = m.id "
                + "JOIN cars_drivers cd ON c.id = cd.car_id "
                + "WHERE cd.driver_id = ?";
```
for better readability
also lets check if car or driver is not deleted

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete car by id " + id,e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
```
Dmytro-Mk, line 114: same, cars with no drivers

Dmytro-Mk (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,200 @@
...
+        String query = "SELECT c.id, model, manufacturer_id, name, country, cd.driver_id "
+                + "FROM cars c "
+                + "JOIN cars_drivers cd "
+                + "ON c.id = cd.car_id "
+                + "JOIN manufacturers m "
+                + "ON c.manufacturer_id = m.id "
+                + "WHERE cd.driver_id = ? AND c.is_deleted = FALSE "
+                + "AND m.is_deleted = FALSE";
```
Dmytro-Mk, line 124:
```suggestion
                + "AND cd.is_deleted = FALSE";
```
we definitely need to check if we didn't remove driver form the car 
and I think we don't need to heck for manufacturer otherwise it's not going to show car only because manufacturer got deleted, and I think it should show car anyway 
But you can leave manufacturer check, it's not important in this task

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,200 @@
...
+            statement.setLong(1, driverId);
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                carsByDriverList.add(getCar(resultSet));
+            }
+            return carsByDriverList;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get list of cars by driver id "
+                    + driverId, e);
+        }
```
Dmytro-Mk, line 137:
```suggestion
        } catch (SQLException e) {
            throw new DataProcessingException("Can't get list of cars by driver id "
                    + driverId, e);
        }
        cars.foreach( -> car.setDrivers(getDrivers()));
        return cars;
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT c.id, model, manufacturer_id, name, country, cd.driver_id "
+                + "FROM cars c "
+                + "JOIN cars_drivers cd "
+                + "ON c.id = cd.car_id "
+                + "JOIN manufacturers m "
+                + "ON c.manufacturer_id = m.id "
+                + "WHERE cd.driver_id = ? AND c.is_deleted = FALSE "
+                + "AND m.is_deleted = FALSE";
+        List<Car> carsByDriverList = new ArrayList<>();
```
Dmytro-Mk, line 125:
```suggestion
        List<Car> cars = new ArrayList<>();
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        return new Car(resultSet.getObject("id", Long.class),
+                resultSet.getString("model"),
+                new Manufacturer(resultSet.getObject("manufacturer_id", Long.class),
+                        resultSet.getString("name"),
+                        resultSet.getString("country")),
+                getDrivers(resultSet.getObject("id", Long.class)));
```
Dmytro-Mk, line 146: image
just remove getDrivers call from this method and call separately in the end of dao methods

Dmytro-Mk (changes requested):

Dmytro-Mk (approved):

KostyaNikiforov (approved):

PR #481, opened 2023-01-25
join 1st commit

OleksandrKostiukevych:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,199 @@
...
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car by id: " + id + ".", e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriversByCarId(id));
+        }
+        return Optional.ofNullable(car);
```
Uka0001, line 61:
```suggestion
if (car != null) {
            car.setDrivers(getDriversByCarId(id));
            return Optional.of(car);
        }
        return Optional.empty();
```

Uka0001 (approved):
Good work, minor issue and you have to fix it before the mentor reviews your work. Some method names I don't like but, I think, it's ok because you can have your own choice and for me, it was clear what they are used for.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,45 @@
...
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver petrenkoPP = new Driver("Petrenko P.P.", "KIA000001");
+        Driver ivanenkoII = new Driver("Ivanenko I.I.", "KIB000001");
+        Driver boykoOO = new Driver("Boyko O.O.", "KIC000001");
+        Driver sydorenkoBB = new Driver("Sydorenko B.B.", "KIE000001");
+        driverService.create(petrenkoPP);
+        driverService.create(ivanenkoII);
+        driverService.create(boykoOO);
+        driverService.create(sydorenkoBB);
```
Uka0001, line 24: you can add them to list and create them in forEach, its proposition in case you have 100 manufacturers/drivers/cars :) but in case of 4 I think its ok :)

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,21 @@ CREATE TABLE `drivers` (
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
+CREATE TABLE `cars_drivers` (
+  `car_id` bigint NOT NULL,
+  `driver_id` bigint NOT NULL,
+  KEY `car_id_idx` (`car_id`),
+  KEY `driver_id_idx` (`driver_id`,`car_id`),
+  CONSTRAINT `car_id` FOREIGN KEY (`car_id`) REFERENCES `cars` (`id`),
+  CONSTRAINT `driver_id` FOREIGN KEY (`driver_id`) REFERENCES `drivers` (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
+
```
KostyaNikiforov, line 38: Redundant empty line

KostyaNikiforov (approved):

PR #482, opened 2023-01-25
CarService and CarDao implemented

taras-zhyravetskyi:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,197 @@
...
+            if (resultSet.next()) {
+                car = parseCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car with id: " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getDrivers(id));
+        }
+        return Optional.ofNullable(car);
```
Uka0001, line 59:
```suggestion
        if (car != null) {
            car.setDrivers(getDrivers(id));
            return Optional.of(car);
        }
        return Optional.empty();
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = new ArrayList<>(car.getDrivers());
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        List<Driver> drivers = new ArrayList<>(car.getDrivers());
```
Uka0001, line 54:
```suggestion
        List<Driver> drivers = car.getDrivers();
```

Uka0001 (approved):
Good work, minor issues or even propositions :)

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
taras-zhyravetskyi: We should create a new ArrayList because List<Drivers> can be immutable. In your suggestion there isn't possibility to add a new driver to the immutable list

Uka0001: ok, I miss this point, because I think only List.of is immutable...

KostyaNikiforov (approved):

PR #490, opened 2023-01-26
Car created

Gleb-Shevchenko:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,199 @@
...
+            throw new DataProcessingException("Couldn't get a list of drivers from driversDB.", e);
+        }
+        for (Car car : cars) {
+            car.setDrivers(getDriversByCar(car.getId()));
+        }
+        return cars;
+    }
+
+    @Override
+    public Car update(Car car) {
```
MaxMaksym, line 89: you need to call here 2 methods: to delete old relations between drivers and cars and to add new one.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,199 @@
...
+        return Optional.ofNullable(car);
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT cars.id AS car_id, model, "
+                + "manufacturers.id AS manufacturer_id, name, country "
+                + "FROM cars "
+                + "JOIN manufacturers ON cars.manufacturer_id = manufacturers.id "
+                + "WHERE cars.is_deleted = FALSE";
```
MaxMaksym, line 71: It's better to use aliases in all your queries, but let it be as it is)

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -7,6 +7,14 @@ public class Driver {
...
+    }
+
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
     public Driver(Long id, String name, String licenseNumber) {
         this.id = id;
         this.name = name;
```
MaxMaksym, line 20: build your logic either on constructors or on setters, 3 overloaded is too much, try to remove at least one of them

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -7,6 +7,14 @@ public class Manufacturer {
...
+    }
+
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
+
     public Manufacturer(Long id, String name, String country) {
         this.id = id;
         this.name = name;
```
MaxMaksym, line 20: same

MaxMaksym (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
Gleb-Shevchenko: It is better now without them. They get me confused a bit

MaxMaksym (approved):

KostyaNikiforov (approved):

PR #500, opened 2023-02-05
tasks were implemented

lysytskapolina:


src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, model, manufacturer, drivers);
+    }
+
+    @Override
+    public String toString() {
+        return "Car{" + "id=" + id + ", model='" + model + '\''
+                + ", manufacturer=" + manufacturer + ", drivers=" + drivers + '}';
```
Serhii1206, line 76: public String toString() {
        return "Car{" + "id=" + id 
                   + ", model='" + model + '\''
                   + ", manufacturer=" + manufacturer 
                   + ", drivers=" + drivers 
                   + '}';

Serhii1206 (changes requested):

src/main/java/mate/jdbc/model/Car.java
lysytskapolina: I can't agree that it's so necessary, but anyways it's done😌

Serhii1206 (approved):
Great job

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,45 @@
...
+                new Car(0L, "Countryman", mini, List.of(roman)),
+                new Car(0L, "Octavia", skoda, drivers)
+        );
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        cars.forEach(carService::create);
+        System.out.println(carService.getAll());
+        System.out.println(carService.getAllByDriver(1L));
+        carService.addDriverToCar(polina, carService.get(2L));
+        carService.removeDriverFromCar(polina, carService.get(3L));
+        carService.delete(1L);
```
KostyaNikiforov, line 42: Id must be generated by the database. Do not hardcode it in the code

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #505, opened 2023-02-09
First commit.

Taurusden:
1. Car class was created.
2. Table cars was created. Created table cars_drivers which connect to BDs (many-to-many).
3. Created all method in service and in Dao.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,26 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.CarService;
+
 public class Main {
+
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        Manufacturer manufacturer = new Manufacturer(2L, "Peugeot", "Hungary");
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        carService.create(new Car("106", manufacturer));
+        carService.get(6L);
+        Car car = carService.update(new Car(6L,"206", manufacturer));
+        carService.delete(9L);
+        carService.removeDriverFromCar(new Driver(2L),new Car(2L));
+        Driver driver = new Driver(1L, "Denis", "123456789");
+        carService.addDriverToCar(driver,car);
+        System.out.println(carService.getAll());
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.List;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+
+public interface CarDao {
+
+    Car create(Car car);
+
+    Car get(Long id);
```
okuzan, line 11: dao should return optional

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.CarService;
+
 public class Main {
+
```
okuzan, line 10: delete empty lines after class declaration

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        Manufacturer manufacturer = new Manufacturer(2L, "Peugeot", "Hungary");
```
okuzan, line 14: don't pass id, id should be generated on saving. Generally, avoid hardcoded values as much as possible, instead of referencing it explicitly, call manufacturer.getId();

okuzan (changes requested):

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,24 @@
...
+
+    Car update(Car car);
+
+    boolean delete(Long id);
+
+    List<Car> getAllByDriver(Long driverId);
+
+    void addDriverToCar(Driver driver, Car car);
+
+    void removeDriverFromCar(Driver driver, Car car);
```
okuzan, line 23: image

okuzan (changes requested):

src/main/java/mate/jdbc/dao/CarDao.java
Ivan95kos: Not fixed: 
Знімок екрана 2023-02-11 о 12 27 13

Ivan95kos (changes requested):
Please, fix previous comment

Ivan95kos (approved):
Great job

PR #510, opened 2023-02-15
All tasks done.

svitlana-tertyshna:
Created class Car, CarService interface amd implementation; tested solution in the mein method.

### DIFF
--- a/src/main/java/mate/jdbc/service/CarServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/CarServiceImpl.java
@@ -1,0 +1,63 @@
+package mate.jdbc.service;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.NoSuchElementException;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+
+@Service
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = new ArrayList<>(car.getDrivers());
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        update(car);
```
aaananas, line 48:
```suggestion
        car.getDrivers().add(driver);
        update(car);
```

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        List<Driver> drivers = new ArrayList<>(car.getDrivers());
+        drivers.remove(driver);
+        car.setDrivers(drivers);
+        update(car);
```
aaananas, line 56:
```suggestion
        car.getDrivers().remove(driver);
        update(car);
```

aaananas (approved):

PR #512, opened 2023-02-16
feat: jv-join #comment was added posability to wok with Car model

sania517:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
```
Ivan95kos, line 19: you need to create an impl package and move the implementation there

Ivan95kos (changes requested):
do not leave the commented code, delete it if it is not needed

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.List;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+
+@Service
+public class CarServiceImpl implements CarService {
```
Ivan95kos, line 11: you need to create an impl package and move the implementation there

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,66 @@
...
+        //Driver daniel = driverService.get(2L);
+        //Driver bill = driverService.get(3L);
+        //3.setDrivers(List.of(daniel, bill));
+        //carService.update(i3);
+        //System.out.println(carService.get(5L));
+
+        //8. testing carService.delete
+        //Car i3 = carService.get(5L);
+        //carService.delete(i3.getId());
+        //System.out.println(carService.getAll());
```
aaananas, line 64: rather remove or uncomment it
because it is needed to test your code you should leave it ucommented

aaananas (changes requested):

kozhukhovsky (approved):

PR #513, opened 2023-02-17
Task is done.

Viper7557:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
+import mate.jdbc.service.CarService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        Car prado = new Car();
+        prado.setModel("Prado");
+        prado.setManufacturer(new Manufacturer(2L, "Toyota", "Japan"));
```
aleksandr-bevza, line 17: Why do you put 'id' into new Manufacturer and Driver? You can see in Dao implementation that this 'id' is skipped. Database generates it's own id.
image

It's better to create new manufacturer using ManufacturerService and then put it into car. Same for new Driver

src/main/java/mate/jdbc/service/CarService.java
```diff
@@ -0,0 +1,24 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+
+public interface CarService {
+
```
aleksandr-bevza, line 8:
```suggestion
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        String manufacturerName = resultSet.getString("name");
+        String manufacturerCountry = resultSet.getString("country");
+        Manufacturer manufacturer =
+                new Manufacturer(manufacturerId, manufacturerName, manufacturerCountry);
+        Car car = new Car();
+        car.setId(id);
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(getDriversForCar(id));
+        insertDrivers(car);
```
aleksandr-bevza, line 147: This method (getCar) should only parse car from result set. Do you need to perform this 'insert drivers' operation here?

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,21 @@ CREATE TABLE `drivers` (
...
+  FOREIGN KEY (`manufacturer_id`)
+  REFERENCES `manufacturers` (`id`)
+  ON DELETE NO ACTION
+  ON UPDATE NO ACTION);
+
+  CREATE TABLE `cars_drivers` (
+  `car_id` BIGINT NOT NULL,
+  `driver_id` BIGINT NOT NULL,
+  CONSTRAINT `cars_drivers_cars_fk` FOREIGN KEY (`car_id`) REFERENCES `cars` (`id`),
+  CONSTRAINT `cars_drivers_drivers_fk` FOREIGN KEY (`driver_id`) REFERENCES `driverscars_drivers` (`id`)
```
aleksandr-bevza, line 37:
```suggestion
  CONSTRAINT `cars_drivers_drivers_fk` FOREIGN KEY (`driver_id`) REFERENCES `drivers` (`id`)
```

aleksandr-bevza (changes requested):

PR #515, opened 2023-02-19
jv-join-solution

nkharyshyn:
-connected project to my db taxi-service;
-created class Car;
-created dao and service layers for class Car;
-in dao layer for Car class added CRUD methods and few additional methods for working with resultSets and creating instances; 
-in service layer rewrote same methods as in dao layer and added 2 only-service-layer-methods, which logic wrote using update method;
-tested my methods in Main class (due to some mistakes and problems I had to create and delete few cars, that's why I used id 19-20 for my cars)

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,65 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Long FIRST_DRIVER_ID = 1L;
+    private static final Long SECOND_DRIVER_ID = 2L;
+    private static final Long THIRD_DRIVER_ID = 3L;
+    private static final Long FIRST_CAR_MANUFACTURER_ID = 2L;
+    private static final Long SECOND_CAR_MANUFACTURER_ID = 7L;
+    private static final Long FIRST_CAR_ID = 17L;
+    private static final Long SECOND_CAR_ID = 18L;
+    private static final String FIRST_CAR_MODEL = "A6";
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        car.setDrivers(drivers);
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        drivers.remove(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
```
aleksandr-bevza, line 61:
```suggestion
        car.getDrivers().remove(driver);
        update(car);
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
```
aleksandr-bevza, line 53:
```suggestion
        car.getDrivers().add(driver);
        update(car);
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,65 @@
...
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Long FIRST_DRIVER_ID = 1L;
```
aleksandr-bevza, line 13: Better not use hardcoded ids.
To test your functionality - create new manufacturers, drivers and cars in method 'main'
Do not use hardcoded ids

aleksandr-bevza (changes requested):

kozhukhovsky (approved):
Good job!

PR #517, opened 2023-02-24
1

dimonkoval:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,19 @@
...
+    Car get(Long id);
+
+    List<Car> getAll();
+
+    Car update(Car car);
+
+    boolean delete(Long id);
+
+    List<Car> getAllByDriver(Long driverId);
+
```
kozhukhovsky, line 18: remove empty line

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            throw new DataProcessingException("Couldn't create "
+                    + car + ". ", e);
+        }
+    }
+
+    @Override
+    public Car get(Long id) {
+        String query = "SELECT id, model, manufacturer_id, driver_id "
+                + "FROM cars c JOIN cars_drivers cd "
+                + "ON c.id = cd.car_id where is_deleted = false AND c.id = ?";
```
kozhukhovsky, line 50:
```suggestion
                + "ON c.id = cd.car_id WHERE is_deleted = FALSE AND c.id = ?";
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create "
+                    + car + ". ", e);
+        }
+    }
+
+    @Override
+    public Car get(Long id) {
```
kozhukhovsky, line 47: Its better to return `Optional<Car>`

src/main/java/mate/jdbc/dao/CarDaoImpl.java
kozhukhovsky: You should join manufacturer here

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars "
+                    + "for driver by id " + driverId, e);
+        }
+    }
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject("id", Long.class);
+        String model = resultSet.getString("model");
+        ManufacturerServiceImpl manufacturerService =
```
kozhukhovsky, line 181: You don't have to call service methods at the Dao level, do the joins directly in the request.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT * FROM cars WHERE is_deleted = FALSE";
```
kozhukhovsky, line 67: need join manufacturers here

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+                    insertDriverToCar(carId, driver.getId());
+                }
+            }
+        } else if (driversForCar.size() < currentDrivers.size()) {
+            for (Driver driver : currentDrivers) {
+                if (!driversForCar.contains(driver)) {
+                    deleteDriverFromCar(carId, driver.getId());
+                }
+            }
+        }
```
kozhukhovsky, line 113: It would be better to put this logic into separate methods and call them, for example: isertDriversToCar(), deleteDriversFromCar()

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT * FROM drivers d "
```
kozhukhovsky, line 159: join manufacturer here too

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+                manufacturerService.get(resultSet.getObject("manufacturer_id", Long.class));
+        DriverServiceImpl driverService =
+                (DriverServiceImpl) injector.getInstance(DriverService.class);
+        List<Driver> drivers = getDriversForCar(id);
+        return new Car(id, model, manufacturer, drivers);
+    }
+
+    private List<Driver> getDriversForCar(Long carId) {
+        String query = "SELECT * FROM drivers d JOIN cars_drivers cd "
+                + "ON d.id = cd.driver_id WHERE cd.car_id = ? AND d.is_deleted = false;";
```
kozhukhovsky, line 193:
```suggestion
                + "ON d.id = cd.driver_id WHERE cd.car_id = ? AND d.is_deleted = FALSE;";
```

kozhukhovsky (changes requested):
Good job! Need small fixes :)

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).isPresent() ? carDao.get(id).get() : null;
```
Ivan95kos, line 23: If there is no such entity, it is better to throw Exception.
Also, here we refer to the database twice, it is not necessary to do this
```suggestion
        return carDao.get(id).orElseThrow(() ->
                new NoSuchElementException("Can't get car by id: " + id));
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,14 @@
...
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.CarServiceImpl;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarServiceImpl carService = (CarServiceImpl) injector.getInstance(CarService.class);
```
Ivan95kos, line 11: All methods should be tested

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,213 @@
...
+
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update "
+                    + car + " in carsDB.", e);
+        }
+        updateListDrivers(carId);
+        return car;
+    }
+
+    private void updateListDrivers(Long carId) {
```
Ivan95kos, line 103: All private methods should be below public methods

Ivan95kos (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,14 @@
...
+        System.out.println(carService.get(6L));
```
Ivan95kos, line 12: Id must be generated by the database. Do not hardcode it in the code

kozhukhovsky (approved):

PR #522, opened 2023-03-04
added car model to taxi service, implemented cardao with its implemen…

YevheniiKilovyi:
…tation and carservice and all tested in main

### DIFF
--- a/src/main/java/mate/jdbc/dao/CarDao.java
+++ b/src/main/java/mate/jdbc/dao/CarDao.java
@@ -1,0 +1,20 @@
+package mate.jdbc.dao;
+
+import java.util.List;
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,20 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+
```
uliana-andrushchuk, line 8:
```suggestion
```

uliana-andrushchuk (approved):

PR #523, opened 2023-03-04
Join solution was implemented

kateryna-mykh:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -14,14 +15,24 @@ public class DriverServiceImpl implements DriverService {
...
-        return driverDao.create(driver);
+        Optional<Driver> driverByLicenseNumber = 
+                driverDao.getByLicenseNumber(driver.getLicenseNumber());
+        if (driverByLicenseNumber == null) {
+            return driverDao.create(driver);
+        } else {
+            /*
+             * I decided that the license number should be unique, so if we already have it,
+             * we should return the existing object
+             */
```
uliana-andrushchuk, line 26: In this case, maybe it would be better to rename this method to `getOrCreate()`

uliana-andrushchuk (changes requested):

olekskov (approved):
Good job!

PR #534, opened 2023-03-17
jv-join solution was created.

mrGvvD:
Added Car class.
Implemented CarServiceImpl and CarDaoImpl.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,65 @@
 package mate.jdbc;
 
+import java.util.ArrayList;
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturerUsa = new Manufacturer(null,"Ford", "USA");
+        Manufacturer manufacturerJapan = new Manufacturer(null,"Toyota", "Japan");
+        Manufacturer manufacturerSweden = new Manufacturer(null,"Volvo", "Sweden");
+        manufacturerService.create(manufacturerUsa);
+        manufacturerService.create(manufacturerJapan);
+        manufacturerService.create(manufacturerSweden);
+
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver john = new Driver(null,"John","12345678");
+        Driver bob = new Driver(null,"Bob","23456789");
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,65 @@
...
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturerUsa = new Manufacturer(null,"Ford", "USA");
```
Ivan95kos, line 20: It is better to create an additional constructor without id
do the same for Driver and Car
```suggestion
        Manufacturer manufacturerUsa = new Manufacturer("Ford", "USA");
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get a list of cars", e);
+        }
+    }
+
+    private void addDrivers(List<Driver> driverList, Long carId) {
+        String query = "INSERT INTO cars_drivers (driver_id, car_id) VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(query,
+                         Statement.RETURN_GENERATED_KEYS)) {
```
Ivan95kos, line 126: You don't use statement.getGeneratedKeys() in this method
```suggestion
                 PreparedStatement statement = connection.prepareStatement(query)) {
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't add drivers: " + carId, e);
+        }
+    }
+
+    private void removeDriverFromCar(Car car) {
+        String query = "DELETE FROM cars_drivers WHERE car_id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(query,
+                         Statement.RETURN_GENERATED_KEYS)) {
```
Ivan95kos, line 141: the same
```suggestion
                 PreparedStatement statement = connection.prepareStatement(query)) {
```

Ivan95kos (changes requested):

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public void setDriver(List<Driver> drivers) {
+        this.driverList = drivers;
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
uliana-andrushchuk, line 63: Your implementation of the `hashCode()` method does not seem to provide any functionality, as it simply returns the hash code of the superclass. 
Please, add custom logic to this method. The same for `equals` and `toString`

uliana-andrushchuk (changes requested):

olekskov (approved):
Good job, approved. Please note for further that it's better to use as narrow joins as possible in your queries. In this task inner join is suitable

PR #537, opened 2023-03-20
Added solution to the task

ivbudenkov:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+                    + car + " in carsDB.", e);
+        }
+        deleteDriverList(car);
+        createDriverList(car);
+        return car;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE cars SET is_deleted = TRUE WHERE id = ?";
```
olekskov, line 109: I'd add also check for isDeleted flag here in order not to return true for repetitive delete request for same id

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+        return cars;
+    }
+
+    private void createDriverList(Car car) {
+        if (car.getDrivers().isEmpty()) {
+            return;
+        }
+        StringBuilder values = new StringBuilder();
+        for (var ignored : car.getDrivers()) {
```
olekskov, line 151: Code style should be consistent. You use particular type everywhere else but here you used var instead

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+        String query = "INSERT INTO cars_drivers (car_id, driver_id) VALUES "
+                + values;
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            int counter = 1;
+            for (Driver driver : car.getDrivers()) {
+                statement.setLong(counter++, car.getId());
+                statement.setLong(counter++, driver.getId());
+            }
+            statement.executeUpdate();
```
olekskov, line 163: With such change it should work properly without StringBuilder usage:
```suggestion
            statement.setLong(1, car.getId());
            for (Driver driver : car.getDrivers()) {
                statement.setLong(2, driver.getId());
                statement.executeUpdate();
            }
```

olekskov (changes requested):

src/main/java/mate/jdbc/dao/CarDaoImpl.java
ivbudenkov: in such case if a car has 10 drivers we`d execute 10 queries to DB instead of one

kozhukhovsky: its ok

kozhukhovsky (approved):

PR #538, opened 2023-03-20
Hw jv join

Iskamele:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't find car by id " + id + ". ", e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriversForCar(id));
+            return Optional.of(car);
+        } else {
+            return Optional.empty();
+        }
```
aaananas, line 64:
```suggestion
        if (car != null) {
            car.setDrivers(getDriversForCar(id));
            return Optional.of(car);
        }
        return Optional.empty();
```

aaananas (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
Iskamele: Yes, this way the code looks more concise and readable

PR #546, opened 2023-03-25
JDBC CarService module finished

IlkoGood:


AnonimToster (approved):
Good solution!)

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,192 @@
...
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection.prepareStatement((query),
```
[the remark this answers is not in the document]
IlkoGood: Well yeah, I've started with this expeirence too, but then I saw default implementations on mate repo. :)

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,95 @@
...
+@Service
+public class CarServiceImpl implements CarService {
+    @Inject
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        if (car == null) {
+            throw new DataProcessingException("Can't create car from null");
+        }
```
karma-o, line 21: lets replace **all** null checks with  [Objects.requireNonNull()](https://stackoverflow.com/questions/45632920/why-should-one-use-objects-requirenonnull), wich is more standardised and compact.

you can also add a JetBrains `NotNull` annotation to the needed parameter in the method signature to make it even more explicit and shown in the hint when you hover your mouse over the method name if you want

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,127 @@
...
-        // test your code here
+        List<Manufacturer> manufacturers = testManufacturerService();
+        List<Driver> drivers = testDriverService();
+        testCarService(manufacturers, drivers);
+        // UNCOMMENT METHOD BELOW WITH CAUTION
+        //clearAll();
+    }
+
+    private static void testCarService(List<Manufacturer> manufacturers, List<Driver> drivers) {
+        System.out.println("\nCAR TEST SERVICE \n");
```
karma-o, line 44: use system line separators instead

karma-o (changes requested):

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
IlkoGood: add anotation sounds the best, thanks!

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,128 @@
...
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+    private static final Long FIRST_ID = 1L;
+    private static final Long SECOND_ID = 2L;
+    private static final Long THIRD_ID = 3L;
+    private static final int FOURTH_DRIVER = 3;
+    private static final int FIRST_DRIVER = 1;
+    private static final int FIRST_MANUFACTURER = 0;
+    private static final int SECOND_MANUFACTURER = 1;
+    private static final String SEP = System.lineSeparator();
```
KseniiaMakarova, line 28:
```suggestion
    private static final String SEPARATOR = System.lineSeparator();
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,128 @@
...
+        List<Driver> drivers = new ArrayList<>();
+        Driver bob = driverService.create(
+                new Driver("Bob", "NY12345678"));
+        drivers.add(bob);
+        Driver john = driverService.create(
+                new Driver("John", "FL12345678"));
+        drivers.add(john);
+        Driver alice = driverService.create(
+                new Driver("Alice", "NE12345678"));
+        System.out.println(driverService.get(SECOND_ID));
```
KseniiaMakarova, line 76:
```suggestion
        System.out.println(driverService.get(john.getId()));
```
same everywhere
please don't use hardcoded ids

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,128 @@
...
+        System.out.println(driverService.get(THIRD_ID));
+        Driver aliceToEmma = driverService.update(
+                new Driver(THIRD_ID, "Emma", "TE12345678"));
+        drivers.add(aliceToEmma);
```
KseniiaMakarova, line 80: wrong understanding of method update()

after creation id is set to your object, and you can use it in code
and you also have object which you can change (set another name, for example)
than you should call update() method with this updated object (and not pass the new object there!)
also, no need to call add() one more time in the end

```suggestion
        alice.setName("Emma");
        Driver aliceToEmma = driverService.update(alice);
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,128 @@
...
+                new Manufacturer("Hyundai", "South Korea"));
+        manufacturers.add(hyundai);
+        Manufacturer cherry = manufacturerService.create(
+                new Manufacturer("Cherry", "China"));
+        System.out.println(manufacturerService.get(SECOND_ID));
+        System.out.println(manufacturerService.get(THIRD_ID));
+        Manufacturer cherryToHaval = manufacturerService.update(
+                new Manufacturer(THIRD_ID, "Haval", "China"));
+        manufacturers.add(cherryToHaval);
+        manufacturerService.delete(SECOND_ID);
```
KseniiaMakarova, line 105:
```suggestion
        manufacturerService.delete(hyundai.getId());
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,128 @@
...
+        Car amarok = carService.create(
+                new Car("Amarok",
+                        manufacturers.get(FIRST_MANUFACTURER),
+                        drivers));
+        Car santaFe = carService.create(
+                new Car("Santa Fe",
+                        manufacturers.get(SECOND_MANUFACTURER),
+                        drivers.stream().limit(FIRST_DRIVER).collect(Collectors.toList())));
+        System.out.println(carService.get(SECOND_ID));
+        carService.addDriverToCar(drivers.get(FOURTH_DRIVER), santaFe);
```
KseniiaMakarova, line 55: looks a bit confusing
introduce local variables for drivers and manufacturers to make code more readable

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
KseniiaMakarova, line 1: for future:
please don't touch files that are already added to project
it is hard to review whole project each time instead of separate parts

you have a project - and you just need to write some new files

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        String query = "INSERT INTO cars_drivers (driver_id, car_id) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection.prepareStatement(query)) {
+            preparedStatement.setLong(2, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                preparedStatement.setLong(1, driver.getId());
+                preparedStatement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't add to DB: " + car, e);
```
KseniiaMakarova, line 142:
```suggestion
            throw new DataProcessingException("Can't add drivers to car: " + car, e);
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    private List<Driver> getDriversForCar(Long carId) {
+        String query = "SELECT d.id AS driver_id, d.name AS name, d.license_number AS license "
```
KseniiaMakarova, line 147: do we really need this aliases?

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get driver for carID " + carId, e);
+        }
+    }
+
+    private Driver getDriverFromResultSet(ResultSet resultSet) throws SQLException {
+        return new Driver(
+                resultSet.getObject("driver_id", Long.class),
+                resultSet.getString("name"),
+                resultSet.getString("license"));
```
KseniiaMakarova, line 168:
```suggestion
                resultSet.getObject("id", Long.class),
                resultSet.getString("name"),
                resultSet.getString("license_number"));
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,196 @@
...
+            preparedStatement.setLong(1, driverId);
+            ResultSet resultSet = preparedStatement.executeQuery();
+            while (resultSet.next()) {
+                carIds.add(resultSet.getObject("car_id", Long.class));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't form a list of cars for driver id "
+                    + driverId, e);
+        }
+        return carIds.stream().map(id -> get(id).get()).collect(Collectors.toList());
```
KseniiaMakarova, line 97: here for 50 cars you will do 51 request
"n+1 problem"

let's use 2 queries like in getAll() method
but add 1 more join in the first query

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public List<Car> getAllByDriver(@NotNull Long driverId) {
+        return carDao.getAllByDriver(driverId);
+    }
+
+    @Override
+    public Car update(@NotNull Car car) {
+        Objects.requireNonNull(car, "car must not be null");
```
KseniiaMakarova, line 45: why do we need both annotation and Objects.requireNonNull?

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import mate.jdbc.service.CarService;
+import org.jetbrains.annotations.NotNull;
+
+@Service
+public class CarServiceImpl implements CarService {
+    @Inject
+    private CarDao carDao;
+
+    @Override
+    public Car create(@NotNull Car car) {
```
KseniiaMakarova, line 21: is it annotation only for compilation (doesn't throw exception in Runitme)? if so, let's remove it
you can use javax NotNull annotation to remove checking for Objects.requireNonNull

KseniiaMakarova (changes requested):

sarakhmen (approved):
Good job ;)

PR #552, opened 2023-03-26
jv-join solution

potrap:


AndrewkeY-hub (approved):
Looks good

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+                car = parseCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car with id: " + id, e);
+        }
+        if (car == null) {
+            return Optional.empty();
+        }
+        setDriversList(car);
+        return Optional.of(car);
```
petro-ushchuk, line 72: U can move lines 71-72 in if block after parsing the car, and return just optional empty after try-catch block. Also, after this, you don't need the local variable at the line 57, u can aslo move it to the if block. It would make your code more readable.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection.prepareStatement(query)) {
+            preparedStatement.setLong(1, driverId);
+            ResultSet resultSet = preparedStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                Long carId = resultSet.getLong("car_id");
+                Car car = get(carId)
+                        .orElseThrow(() -> new DataProcessingException("Couldn't find car"
+                                + " with such id: " + carId));
```
petro-ushchuk, line 139: refactor it to use only 1 sql query instead of calling `get` method n-times.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+                Long driverId = resultSet.getLong("driver_id");
+                Driver driver = driverDao.get(driverId)
+                        .orElseThrow(() -> new DataProcessingException("No driver with such id: "
+                        + driverId));
+                drivers.add(driver);
+            }
+            car.setDrivers(drivers);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get drivers for car: " + car, e);
+        }
```
petro-ushchuk, line 192: the same here, you don't need to call n-times `get` of `DriverDao`, write only 1 query.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private Car parseCar(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        car.setId(resultSet.getLong("car_id"));
+        car.setModel(resultSet.getString("model"));
+        car.setManufacturer(manufacturerDao.get(resultSet.getLong("manufacturer_id"))
+                .orElseThrow(() -> new DataProcessingException("Couldn't parse the car")));
```
petro-ushchuk, line 200: u can retrive the manufacturer by join to the table and parse the resultSet. please, don't call this method, u should use joins

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,31 @@ CREATE TABLE `drivers` (
...
+    REFERENCES `taxi_service_db`.`drivers` (`id`)
+    ON DELETE NO ACTION
+    ON UPDATE NO ACTION,
+  CONSTRAINT `car_id`
+    FOREIGN KEY (`car_id`)
+    REFERENCES `taxi_service_db`.`cars` (`id`)
+    ON DELETE NO ACTION
+    ON UPDATE NO ACTION);
+
+
```
petro-ushchuk, line 47: what should we do with redundant empty lines?

petro-ushchuk (changes requested):

sarakhmen (approved):
Good job ;)

PR #555, opened 2023-03-27
Implemented Car Service

De1eF:


Nikitos787 (approved):

outref (changes requested):
Cool soulution) 
Just some ideas and suggestions from me

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars "
+                    + "from cars table. ", e);
+        }
+        for (Car car : cars) {
+            if (car != null) {
+                car.setDrivers(getDriversForCar(car.getId()));
+            }
+        }
```
sarakhmen, line 97: I'd recommend using forEach method here instead of loop. Also, is it possible for a car to be null?

sarakhmen (changes requested):
Overall good job, just a minor fixes required

karma-o (approved):

PR #561, opened 2023-03-27
implement carService and carDao layers

denys-mg:


VitaliiPotseluiko (changes requested):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,224 @@
...
+            deleteCarStatement.setLong(1, id);
+            return deleteCarStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't execute statement"
+                    + " to delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getCarsForDriver(Long driverId) {
```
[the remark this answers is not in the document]
denys-mg: Yeah, but it's also not so bad name)

VitaliiPotseluiko (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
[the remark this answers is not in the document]
karma-o: let's stick to the one we have in the task description

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+                insertCarAndDriverStatement.setLong(2, driver.getId());
+                insertCarAndDriverStatement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't execute statement "
+                    + "to insert car and it's drivers relationship" + car, e);
+        }
+    }
+
+    private void deleteCarAndDriver(Car car) {
```
karma-o, line 195: I do not see any car insert or delete in those methods so let's not name them like they are
```suggestion
    private void insertDriversForCar(Car car) {
        String createQuery = "INSERT INTO cars_drivers (car_id, driver_id) VALUES(?, ?);";
        try (Connection connection = ConnectionUtil.getConnection();
                PreparedStatement insertCarAndDriverStatement =
                        connection.prepareStatement(createQuery)) {
            insertCarAndDriverStatement.setLong(1, car.getId());
            for (Driver driver : car.getDrivers()) {
                insertCarAndDriverStatement.setLong(2, driver.getId());
                insertCarAndDriverStatement.executeUpdate();
            }
        } catch (SQLException e) {
            throw new DataProcessingException("Can't execute statement "
                    + "to insert car and it's drivers relationship" + car, e);
        }
    }

    private void deleteDriversOfCar(Car car) {
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    public List<Car> getCarsForDriver(Long driverId) {
+        return carDao.getCarsForDriver(driverId);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers()
+                .stream()
+                .filter(d -> d.getId().equals(driver.getId()))
+                .toArray().length > 0) {
```
karma-o, line 59: you can use `anyMatch()` terminal operation here

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers()
+                .stream()
+                .filter(d -> d.getId().equals(driver.getId()))
+                .toArray().length == 0) {
+            throw new DataProcessingException("There is no such driver with driverId - "
+                    + driver.getId() + " in the car " + car);
+        }
```
karma-o, line 75: this one can be removed because I don't think it is necessary to throw an exception in case of the already absent driver for that car, so I think we can just have a return in this `if`s body, since there is nothing else to do

src/main/resources/init_db.sql
```diff
@@ -1,20 +1,39 @@
...
-DROP TABLE IF EXISTS `drivers`;
-CREATE TABLE `drivers` (
-                                  `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
-                                  `name` VARCHAR(225) NOT NULL,
-                                  `license_number` VARCHAR(225) NOT NULL,
-                                  `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                                  PRIMARY KEY (`id`),
-                                  UNIQUE INDEX `id_UNIQUE` (id ASC) VISIBLE,
-                                  UNIQUE INDEX `license_number_UNIQUE` (`license_number` ASC) VISIBLE);
+DROP TABLE IF EXISTS `taxi_service`.`drivers`;
```
karma-o, line 12: redundant explicit schema call, and the same for others. 
Overall I do not recommend changing the SQL statements/scripts if they were already used at some point, you can have checksums mismatch in the future, when you will be working with RDBMS, but that is a topic for the future
```suggestion
DROP TABLE IF EXISTS `drivers`;
```

karma-o (changes requested):

karma-o (approved):

PR #562, opened 2023-03-27
Implemented CarService service and CarDao class.

VitaliiPotseluiko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,40 @@
...
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+    private static DriverService driverService =
+            (DriverServiceImpl) injector.getInstance(DriverService.class);
+    private static CarService carService = (CarServiceImpl) injector.getInstance(CarService.class);
+
     public static void main(String[] args) {
-        // test your code here
+        Car car = new Car("RenoA2",new Manufacturer(3L,"reno", "France"));
```
denys-mg, line 19: I think it's not a good idea to set id manually, after creation your objects have id

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (manufacturer_id, model) VALUES (?, ?)";
```
denys-mg, line 23: Don't forget to add a semicolon symbol at the end of query

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        String query = "UPDATE cars "
+                + "SET manufacturer_id = ?, model = ? "
+                + "WHERE id = ? AND is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                    PreparedStatement updateStatement = connection.prepareStatement(query)) {
+            updateStatement.setLong(1, car.getManufacturer().getId());
+            updateStatement.setString(2, car.getModel());
+            updateStatement.setLong(3, car.getId());
+            deleteDrivers(car.getId());
+            addDrivers(car);
```
denys-mg, line 92: Move deleteDrivers and addDrivers from try block (as you create connection into another connection)

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,194 @@
...
+            return updateStatement.executeUpdate() > 0 ? car : null;
```
denys-mg, line 93: Maybe it's better to throw here exception in that case instead of passing null value

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,40 @@
...
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.impl.CarServiceImpl;
+import mate.jdbc.service.impl.DriverServiceImpl;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+    private static DriverService driverService =
+            (DriverServiceImpl) injector.getInstance(DriverService.class);
+    private static CarService carService = (CarServiceImpl) injector.getInstance(CarService.class);
```
denys-mg, line 16: Make it final

denys-mg (changes requested):
Good job, left you my suggestions

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).orElseThrow();
```
nacenik, line 23: Throw exception with message

nacenik (changes requested):

nacenik (approved):
LGTM!

PR #563, opened 2023-03-27
implemented CarService and tested it in Main.class

OlehSarapuk:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    public List<Car> getAllByDriver(Long driverId) {
+        String getAllByDriverQuery = "SELECT * FROM cars_drivers WHERE driver_id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllByDriverStatement =
+                        connection.prepareStatement(getAllByDriverQuery)) {
+            getAllByDriverStatement.setLong(1, driverId);
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                cars.add(get(resultSet.getObject("car_id", Long.class)).get());
```
Deschna, line 129: Im not sure that we should open connection inside the other connection

Deschna (approved):
Overall good job

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,206 @@
...
+            throw new DataProcessingException("can't create such car: " + car, e);
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String getQuery = "SELECT c.id AS car_id, model, c.is_deleted AS car_is_deleted, "
+                + "manufacturer_id, name, country, m.is_deleted AS manufacturer_is_deleted "
```
sarakhmen, line 44: do you really need to get these is_deleted columns here? Fix in all places

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,206 @@
...
+                        connection.prepareStatement(deleteQuery)) {
+            deleteStatement.setLong(1, id);
+            return deleteStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete car with such id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
```
sarakhmen, line 121: In this method let's get cars instead of their ids. And then just call getDrivers() to set drivers for each car.

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        List<Car> cars = new ArrayList<>();
+        for (Long id : carIds) {
+            cars.add(get(id).orElseThrow(() ->
+                    new NoSuchElementException("can't get car with such id: " + id)));
+        }
+        return cars;
+    }
+
+    private void insertDrivers(Car car) {
```
sarakhmen, line 144: here I'd recommend to check if drivers are empty first

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getDriverStatement =
+                        connection.prepareStatement(getDriverQuery)) {
+            getDriverStatement.setLong(1, id);
+            ResultSet resultSet = getDriverStatement.executeQuery();
+            while (resultSet.next()) {
+                Driver driver = new Driver();
+                driver.setId(resultSet.getObject("id", Long.class));
+                driver.setName(resultSet.getString("name"));
+                driver.setLicenseNumber(resultSet.getString("license_number"));
```
sarakhmen, line 173: I'd recommend creating a new private method for driver parsing/creation

sarakhmen (changes requested):
Overall good job but let's improve your solution :)

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all info about cars with such driver id: "
+                    + driverId, e);
+        }
+        cars.forEach(car -> car.setDrivers(getDrivers(car.getId())));
+        return cars;
+    }
+
+    private void insertDrivers(Car car) {
+        if (car.getDrivers().isEmpty()) {
```
nacenik, line 140:
```suggestion
        if (!car.getDrivers().isEmpty()) {
```

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #574, opened 2023-03-28
created Car, CarDao, CarDaoImpl and CarService, CarServiceImpl interf…

pechivo:
…aces and classes

### DIFF
--- a/src/main/java/mate/jdbc/dao/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/CarDaoImpl.java
@@ -1,0 +1,214 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+
+    @Override
+    public Car create(Car car) {
+        String creatCarQuery = "INSERT INTO cars (model, manufacturer_id) "
+                + "VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createCarStatement =
+                        connection.prepareStatement(creatCarQuery,
+                             PreparedStatement.RETURN_GENERATED_KEYS)) {
+            createCarStatement.setString(1, car.getModel());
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        car.getDrivers().add(driver);
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
```
EliSoll, line 49: in this method I checked if driver is present in this car before deleting, but I'am not sure that it is obligatory or not - so I just informed you )

EliSoll (approved):
good job)

src/main/java/mate/jdbc/service/CarServiceImpl.java
pechivo: resonable❤️

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,214 @@
...
+            return deletedCarsAmount > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete car with id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        List<Long> carsId = new ArrayList<>();
+        String getAllByDriver = "SELECT car_id FROM cars_drivers WHERE driver_id = ?;";
```
karma-o, line 127: let's not split a car and manufacturer data fetch into 2 parts, and do this all at once using joins and filtering by `driverId`
and then fetch their drivers similar to the `getAll()` method

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        if (car.getDrivers() != null) {
+            car.getDrivers().remove(driver);
+            carDao.update(car);
+        } else {
+            throw new DataProcessingException("This car " + car
+                    + " doesn't have such driver " + driver);
+        }
```
karma-o, line 56: just a better practice of `if`s
```suggestion
        if (car.getDrivers() == null) {
            throw new DataProcessingException("This car " + car
                    + " has null drivers list");
        }
        car.getDrivers().remove(driver);
        carDao.update(car);
```

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -48,12 +48,11 @@ public void addDriverToCar(Driver driver, Car car) {
     @Override
     public void removeDriverFromCar(Driver driver, Car car) {
         if (car.getDrivers() != null) {
```
karma-o, line 50: missed the change in the condition, it should be == null, now that we throw an exception in its body

karma-o (approved):

PR #578, opened 2023-03-28
Jv join practice

AleksandrLts:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,241 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get car by id " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriversForCar(id));
+            return Optional.of(car);
+        }
+        return Optional.empty();
+    }
```
redmanO-o, line 73:
```suggestion
        if (car != null) {
            car.setDrivers(getDriversForCar(id));
        }
        return Optional.ofNullable(car);
    }
    
    
    you can simplify this
```

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,241 @@
...
+        return cars;
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String getAllByDriverRequest = "SELECT c.id, "
+                + "c.model AS model, "
+                + "m.id AS manufacturer_id, "
+                + "m.name AS name, "
+                + "m.country AS country "
```
redmanO-o, line 104:
```suggestion
                + "c.model, "
                + "m.id AS manufacturer_id, "
                + "m.name, "
                + "m.country "
```
in my opinion it may be so

redmanO-o (approved):
Good job!!!
I left a few comments where, in my opinion, the code can be simplified

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,96 @@
 package mate.jdbc;
```
Andriy-dz, line 1: non-informative commit names

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,96 @@
...
+    private static final Manufacturer teslaManufacturer = new Manufacturer("Tesla", "USA");
+
+    private static final Driver carl = new Driver("Carl", "KGN62178MJI0");
+    private static final Driver tommy = new Driver("Tommy", "JKG64573LOP1");
+    private static final Driver john = new Driver("John", "RET43508MHI3");
+    private static final Driver mike = new Driver("Mike", "DFB63211MNI6");
+
+    private static final Car ford = new Car("Fusion", fordManufacturer, new ArrayList<>());
+    private static final Car audi = new Car("A4", audiManufacturer, new ArrayList<>());
+    private static final Car tesla = new Car("ModelY", teslaManufacturer, new ArrayList<>());
```
Andriy-dz, line 37: it`s must be local variable

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,237 @@
...
+    private static final String CAR_MANUFACTURER_ID_COLUMN = "manufacturer_id";
+    private static final String MANUFACTURER_NAME_COLUMN = "name";
+    private static final String MANUFACTURER_COUNTRY_COLUMN = "country";
+    private static final String DRIVER_ID_COLUMN = "id";
+    private static final String DRIVER_NAME_COLUMN = "name";
+    private static final String DRIVER_LICENSE_NUMBER_COLUMN = "license_number";
+
+    @Override
+    public Car create(Car car) {
+        String createCarRequest = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?);";
```
Andriy-dz, line 31:
```suggestion
        String createCarQuery= "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?);";
```
likewise in the following cases

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,237 @@
...
+                        connection.prepareStatement(insertCarIdToCarsDriversRequest)) {
+            insertCarIdToCarDriversStatement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                insertCarIdToCarDriversStatement.setLong(2, driver.getId());
+                insertCarIdToCarDriversStatement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert drivers " + car.getDrivers(), e);
+        }
+    }
```
Andriy-dz, line 236: what is the difference between this method and insertDrivers()

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,58 @@
+package mate.jdbc.service;
```
Andriy-dz, line 1: it is better to move this class to the impl package

petro-ushchuk (changes requested):
Good work, but let's improve the solution

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                    + " from cars_drivers table", e);
+        }
+    }
+
+    private List<Car> addAndSetDriversToCar(ResultSet resultSet) throws SQLException {
+        List<Car> cars = new ArrayList<>();
+        while (resultSet.next()) {
+            cars.add(getCarFromResultSet(resultSet));
+        }
+        cars.forEach(c -> c.setDrivers(getDriversForCar(c.getId())));
```
karma-o, line 217: you now have a nested connection, which we should avoid as it was said in the common mistakes, you can ask @petro-ushchuk later if he meant something specific by the common logic of these methods, but I think he was generally asking about the possibility of a slight code duplication reduction, IMHO we don't have it, so I'd remove this method back as it was

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,221 @@
...
+    private void insertDrivers(Car car) {
+        String insertDriversQuery = "INSERT INTO cars_drivers (car_id, driver_id) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement insertDriversToCarStatement =
+                        connection.prepareStatement(insertDriversQuery)) {
+            insertDriversToCarStatement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                insertDriversToCarStatement.setLong(2, driver.getId());
+                insertDriversToCarStatement.executeUpdate();
+            }
```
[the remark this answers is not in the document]
karma-o: @AleksandrLts what about this one? Have you tried to implement this advice? Or talked about it somewhere else? would be nice if you left some comments here on why it was not done, 
I am not sure if we have it in our lectures, I think we do not, so I will not count it as a mistake, but in case you are still interested, you can check [this](https://www.baeldung.com/jdbc-batch-processing) out and try to fix it

karma-o (approved):

PR #580, opened 2023-03-28
Hw 22 solution

redmanO-o:


AleksandrLts (approved):
Good solution!

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                + "WHERE c.is_deleted = FALSE AND cd.driver_id = ?";
+        List<Car> cars = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, driverId);
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(parseCarWithManufacturerFromResultSet(resultSet));
+            }
+        } catch (SQLException e) {
```
petro-ushchuk, line 112: `getAll` and `getAllByDriver` looks identical, can we extract to separate method the same logic?

petro-ushchuk (approved):
Good work 🔥
Left comment about improvement

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
redmanO-o: did it =)

PR #594, opened 2023-04-13
initial commit

Rolik-Oleksandr:


src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,90 @@
...
+        return Objects.hash(id, model, manufacturer, drivers);
+    }
+
+    public void addDriver(Driver driver) {
+        drivers.add(driver);
+    }
+
+    public void removeDriver(Driver driver) {
+        drivers.remove(driver);
+    }
```
uliana-andrushchuk, line 89: Do you use this methods?

uliana-andrushchuk (changes requested):

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,20 @@
...
+    Optional<Car> get(Long id);
+
+    List<Car> getAll();
+
+    Car update(Car car);
+
+    boolean delete(Long id);
+
+    List<Car> getAllByDriver(Long driverId);
+
```
Ivan95kos, line 19: For the future, remove the last empty line in all classes/interfaces/methods where it is present
```suggestion
```

PR #595, opened 2023-04-18
Solved

arthursilver111:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
okuzan, line 9: Dao should return Option on get method, adjust your implementation

okuzan (changes requested):

kozhukhovsky (approved):

PR #598, opened 2023-04-29
added proper solution for this task

polikarpov13:


src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,190 @@
...
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can not delete a car! ", e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "select id, model, manufacturer_id from cars c "
+                + "join cars_drivers cd on c.id = cd.car_id where cd.driver_id = ? "
```
olekskov, line 110: Please use UPPERCASE for SQL key word in all queries to make them more readable

src/main/java/mate/jdbc/service/CarServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can not dd driver to the car! ", e);
+        }
+        car.getDrivers().add(driver);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        String query = "delete from cars_drivers where car_id = ? and driver_id = ?";
```
olekskov, line 61: Operations with DB should be on dao layer.
Requirement "These methods should be only on the service layer" means that you should reuse 'update car' logic from dao layer to perform required actions

src/main/java/mate/jdbc/dao/CarDaoImpl.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public List<Car> getAll() {
+        String query = "select * from cars where is_deleted = false;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            List<Car> cars = new ArrayList<>();
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(get(resultSet.getLong("id")));
+            }
```
olekskov, line 70: That is not how method should work. You make 1 sql query to select all cars and then make 1 additional query to DB for each car found.
Please make 1 proper query to get all required fields and reuse your parseCarWithManufacturer method in cycle here to retrieve all models from 1 result set

olekskov (changes requested):

olekskov (approved):
Good job

PR #609, opened 2023-05-23
home-work/jv-join

andrii-subash:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,77 @@
...
+
+        savedMercedesBenz.setModel("Benz-GLS");
+        Car updatedMercedesBenz = carService.update(savedMercedesBenz);
+        Car bmvX5FromDB = carService.get(savedBmvX5.getId());
+        carService.delete(savedToyotaCamry.getId());
+        carService.removeDriverFromCar(updatedJohn, bmvX5);
+        carService.addDriverToCar(updatedJohn, mercedesBenz);
+
+        System.out.println(carService.getAllByDriver(updatedJohn.getId()));
+        System.out.println(carService.getAll());
```
okuzan, line 75: thorough testing 👏

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        if (drivers.contains(driver)) {
+            drivers.remove(driver);
+            car.setDrivers(drivers);
+            carDao.update(car);
```
okuzan, line 63: same here

okuzan (approved):
Nice job!

PR #614, opened 2023-05-28
Jv join solution

DenisKalyakin:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,19 @@
...
+
+    Optional<Car> get(Long id);
+
+    List<Car> getAll();
+
+    Car update(Car car);
+
+    boolean delete(Long id);
+
+    List<Car> getAllByDriver(Long driverId); // pid pytannyam...
```
Rommelua, line 18: PR is not the correct place for questions. We have Slack for it)
```suggestion
    List<Car> getAllByDriver(Long driverId);
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (!car.getDrivers().contains(driver)) {
+            car.getDrivers().add(driver);
+            carDao.update(car);
+        } else {
+            throw new DataProcessingException("Couldn't add. "
```
Rommelua, line 50: DataProcessingException is for DAO layer only. 
Here you can remove **else** block. Program behavior is not specified in the task for the case when this Driver is already added. So, you can do nothing.

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers().contains(driver)) {
+            car.getDrivers().remove(driver);
+            carDao.update(car);
+        } else {
+            throw new DataProcessingException("Couldn't remove."
```
Rommelua, line 62: same

Rommelua (changes requested):

src/main/java/mate/jdbc/dao/CarDao.java
DenisKalyakin: My bad!) It was only for me, forgot to delete)

Rommelua (approved):
Good!

aaananas (approved):

PR #617, opened 2023-05-31
task solved

MykytaKat:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Manufacturer manufacturer = new Manufacturer(null, "BMW", "USA");
+        manufacturerService.create(manufacturer);
+        Driver firstDriver = new Driver(null, "Alex", "123");
+        Driver seconfDriver = new Driver(null, "Pavel", "456");
```
sZlishchev, line 23: Just create an overloaded constructor for this purpose

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT cars.id AS car_id, model, "
+                + "name, country, manufacturers.id AS manufacturer_id "
+                + "FROM cars "
+                + "JOIN manufacturers "
+                + "ON cars.manufacturer_id = manufacturers.id "
+                + "WHERE cars.id = ? AND cars.is_deleted = FALSE";
```
sZlishchev, line 49: According to common mistakes, a good practice use aliases like cars c and manufacturers m

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT cars.id AS car_id, model, "
+                + "name, country, manufacturers.id AS manufacturer_id "
+                + "FROM cars "
+                + "JOIN manufacturers "
+                + "ON cars.manufacturer_id = manufacturers.id "
+                + "WHERE cars.is_deleted = FALSE";
```
sZlishchev, line 74: Use aliases here too

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String query = "SELECT car.id AS car_id, model, "
+                + "name, country, manufacturer.id AS manufacturer_id "
+                + "FROM cars car "
+                + "JOIN cars_drivers cd "
+                + "ON car.id = cd.car_id "
+                + "JOIN manufacturers manufacturer "
+                + "ON car.manufacturer_id = manufacturer.id "
+                + "WHERE car.is_deleted = FALSE AND cd.driver_id = ?;";
```
sZlishchev, line 132: And aliases here

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private void insertDrivers(Car car) {
+        String insertRequest = "INSERT INTO cars_drivers (driver_id, car_id)"
+                + "VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(insertRequest,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            for (Driver driver : car.getDrivers()) {
+                statement.setLong(1, car.getId());
+                statement.setLong(2, driver.getId());
```
sZlishchev, line 160: In my opinion, it's better to set carId before for each loop

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't insert driver to data base", e);
+        }
+
+    }
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        Manufacturer manufacturer = new Manufacturer(resultSet.getLong("manufacturer_id"),
+                resultSet.getString("name"), resultSet.getString("country"));
```
sZlishchev, line 171: I'd recommend you create a separate method for parsing manufacturer

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+        return car;
+    }
+
+    private List<Driver> getDriversByCar(Long id) {
+        List<Driver> drivers = new ArrayList<>();
+        String query = "SELECT id, name, license_number "
+                + "FROM cars_drivers "
+                + "JOIN drivers "
+                + "ON cars_drivers.driver_id = drivers.id "
+                + "WHERE car_id = ?";
```
sZlishchev, line 184: Also aliases

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+                        resultSet.getString("license_number"));
+                drivers.add(driver);
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get drivers by car id. " + id, e);
+        }
+    }
+
+    private void updateDrivers(Car car) {
```
sZlishchev, line 202:
```suggestion

    private void insertDrivers(Car car) {
```

sZlishchev: I suppose it's a more informative name

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).get();
```
sZlishchev, line 23:
```suggestion
    public Car get(Long id) {
        return carDao.get(id).orElseThrow(() ->
                new NoSuchElementException("Could not get car. ID:" + id));
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        car.getDrivers().add(driver);
+        update(car);
```
sZlishchev, line 44: I'd recommend you check if a driver already exists don't call the update method and throw an exception

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        car.getDrivers().remove(driver);
+        update(car);
```
sZlishchev, line 50: The update method would be useless if the car doesn't have such a driver

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,22 @@ CREATE TABLE `drivers` (
                                   PRIMARY KEY (`id`),
                                   UNIQUE INDEX `id_UNIQUE` (id ASC) VISIBLE,
                                   UNIQUE INDEX `license_number_UNIQUE` (`license_number` ASC) VISIBLE);
+
+DROP TABLE IF EXISTS `cars`;
+CREATE TABLE `cars` (
+                        `id` bigint NOT NULL AUTO_INCREMENT,
+                        `model` varchar(45) NOT NULL,
+                        `manufacturer_id` bigint NOT NULL,
+                        `is_deleted` tinyint NOT NULL DEFAULT '0',
```
sZlishchev, line 27: Please write variables in uppercase for example bigint -> BIGINT

sZlishchev: And pay attention here you should have a query for creating the whole schema

sZlishchev (changes requested):
Good job! Let's improve your solution))

sZlishchev (changes requested):
Great! Only a couple of comments left))

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                         Statement.RETURN_GENERATED_KEYS)) {
+            statement.setString(1, car.getModel());
+            statement.setLong(2, car.getManufacturer().getId());
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+
+        } catch (SQLException e) {
```
sZlishchev, line 35:
```suggestion
        } catch (SQLException e) {
```

sZlishchev (approved):
Good job! Take a look at the last small suggestion))

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService = (ManufacturerService)
```
maxline, line 16: Let's move this manufacturerService variable to the class level. 
Please do the same for all service variables that generated with using injector.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        driverService.create(seconfDriver);
+        System.out.println(driverService.getAll());
+
+        Car car = new Car();
+        car.setModel("Mercedes");
+        car.setManufacturer(manufacturer);
+        car.setDrivers(List.of(firstDriver, seconfDriver));
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        car.setModel("Skoda");
+        car = carService.get(1L);
```
maxline, line 34: Let's try to not hardcode id values in the code, because it can be changed.
```suggestion
        car = carService.get(car.getId());
```
Please fix other cases in the code.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                 PreparedStatement statement = connection.prepareStatement(query,
+                         Statement.RETURN_GENERATED_KEYS)) {
+            statement.setString(1, car.getModel());
+            statement.setLong(2, car.getManufacturer().getId());
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+
```
maxline, line 34:
```suggestion
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement
+                         = connection.prepareStatement(insertRequest)) {
+            statement.setLong(1, car.getId());
+            for (Driver driver : car.getDrivers()) {
+                statement.setLong(2, driver.getId());
+                statement.executeUpdate();
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't insert driver to data base", e);
```
maxline, line 166: add details about car to the exception message.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+
```
maxline, line 168:
```suggestion
```

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+
+public class Car {
+    private Long id;
+    private String model;
+    private Manufacturer manufacturer;
+    private List<Driver> drivers;
+
+    public Car() {
+
```
maxline, line 13:
```suggestion
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
maxline: @sZlishchev exception is not needed in this case because this is not an error. User of application can don't know if the driver present or not when adding new driver, and program should not show any error except it was not mentioned in business requirements.

src/main/java/mate/jdbc/Main.java
```diff
@@ -11,32 +11,32 @@
...
 public class Main {
     private static final Injector injector = Injector.getInstance("mate.jdbc");
+    private static final ManufacturerService manufacturerService
+            = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private static final DriverService driverService
+            = (DriverService) injector.getInstance(DriverService.class);
+    private static final CarService carService
+            = (CarService) injector.getInstance(CarService.class);
+    private static final Manufacturer manufacturer = new Manufacturer("BMW", "USA");
+    private static Car car = new Car();
```
maxline, line 21: Why car variable is static? please move it to the method main()

src/main/java/mate/jdbc/Main.java
```diff
@@ -11,32 +11,32 @@
...
         manufacturerService.create(manufacturer);
         Driver firstDriver = new Driver("Alex", "123");
         Driver seconfDriver = new Driver("Pavel", "456");
         driverService.create(firstDriver);
         driverService.create(seconfDriver);
         System.out.println(driverService.getAll());
-
-        Car car = new Car();
         car.setModel("Mercedes");
         car.setManufacturer(manufacturer);
```
maxline: You can use constructor with parameters instead of default constructor with setters.
Please fix similar cases if needed.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
maxline: redundant empty line not fixed

maxline (changes requested):
Good job! Left just a couple of minor comments.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -168,11 +168,9 @@ private void insertDrivers(Car car) {
     }
 
     private Car getCar(ResultSet resultSet) throws SQLException {
-        Car car = new Car();
         Manufacturer manufacturer = getManufacturer(resultSet);
+        Car car = new Car(resultSet.getString("model"), manufacturer, Collections.emptyList());
```
maxline, line 172: pleas add also additional a constructor without drivers as parameter, so you don't need pass an empty list here.

maxline (approved):
Well done! Approved

PR #620, opened 2023-06-07
Jv join solution

Ivan-Yuriv:


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).orElseThrow();
```
aaananas, line 23: better throw an exception with your custom exception

aaananas (approved):

PR #621, opened 2023-06-07
solution for jv-join task

vgladkyi:


overpathz (changes requested):
There is nothing here.
Please, push required changes again

PR #623, opened 2023-06-08
done all task

vrnvv:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,25 @@
...
+
+    List<Car> getAll();
+
+    Car update(Car car);
+
+    boolean delete(Long id);
+
+    void addDriverToCar(Driver driver, Car car);
+
+    void removeDriverFromCar(Driver driver, Car car);
```
okuzan, line 22: image

okuzan (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

PR #626, opened 2023-06-11
task completed

marimarig:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,198 @@
...
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String selectQuery = "SELECT c.id as car_id, model, "
+                + "m.id as manufacturer_id, m.name, "
+                + "m.country FROM cars c "
+                + "JOIN manufacturers m ON c.manufacturer_id = m.id "
+                + "WHERE c.id = ? AND c.is_deleted = FALSE;";
```
Tupadzaky, line 50: "as" should be in uppercase

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,198 @@
...
+        return Optional.ofNullable(car);
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT c.id as car_id, model, "
+                + "m.id as manufacturer_id, name, "
+                + "country FROM cars c "
+                + "JOIN manufacturers m ON c.manufacturer_id = m.id "
+                + "WHERE c.is_deleted = FALSE;";
```
Tupadzaky, line 74: "as" should be in uppercase

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,198 @@
...
+                        = connection.prepareStatement(updateQuery)) {
+            statement.setString(1, car.getModel());
+            statement.setLong(2, car.getManufacturer().getId());
+            statement.setLong(3, car.getId());
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update car: "
+                    + car, e);
+        }
+        return null;
```
Tupadzaky, line 105: You should return the updated car here
```suggestion
        return car;
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,198 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars "
+                    + "from cars table.", e);
+        }
+        return cars;
+    }
+
+    @Override
+    public Car update(Car car) {
```
Tupadzaky, line 91: The problem with current implementation of this method, is that it doesn't handle the case when your car has updated list of drivers

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,50 @@
...
+        manufacturerService.getAll().forEach(System.out::println);
+
+        Driver firstDriver = new Driver("Sam", "466574AS");
+        driverService.create(firstDriver);
+        Driver secondDriver = new Driver("Jon", "87dd788");
+        driverService.create(secondDriver);
+        driverService.getAll().forEach(System.out::println);
+
+        Car car = new Car("GLE2hCarDao011", mersedes, List.of(firstDriver, secondDriver));
+        Car car2 = new Car("RX222", toyota, List.of(firstDriver));
```
Tupadzaky, line 35: List.of() will create immutable lists, and because of that you would need to make some workarounds in order for addDriverToCar() and removeDriverFromCar() mehtods to work. So consider using new ArrayList<>() here.

Tupadzaky (changes requested):
Good job overall. But there are some problems with update() and main() methods. Check them  in the comments

Tupadzaky (approved):
Looks good to me

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private void deleteDrivers(Long carId) {
+        String deleteQuery = "DELETE FROM cars_drivers WHERE car_id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(deleteQuery)) {
+            statement.setLong(1, carId);
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete relations from DB for car"
+                    + " and drivers with car id: " + carId, e);
```
rovein, line 168: And good here.

PR #627, opened 2023-06-11
My branch

V-Solianyk:


Tupadzaky (approved):
Talk about your solution with mentors. For me it looks very nice, but I don't know if it is actually valid.

src/main/java/mate/jdbc/dao/impl/CarDriverDaoImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import mate.jdbc.dao.CarDriverDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.CarDriver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDriverDaoImpl implements CarDriverDao {
```
maxline, line 18: This class is redundant. Move the needed code to the `CarDaoImpl` class

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+import mate.jdbc.model.CarDriver;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.CarService;
+
+@Service
+public class CarServiceImpl implements CarService {
+    @Inject
+    private CarDao carDao;
+    @Inject
+    private CarDriverDao carDriverDao;
```
maxline, line 19: remove `carDriverDao` field and refactor methods to use only `carDao`

maxline (changes requested):
Good job! Let's improve the solution

rovein (approved):
Great job!

PR #643, opened 2023-06-14
done join task

ViktoriiaSemenchenko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,54 @@
...
+        Driver driverIvan = new Driver("Ivan", "1234");
+        Driver driverVasya = new Driver("Vasya", "5678");
+
+        driverService.create(driverIvan);
+        driverService.create(driverVasya);
+
+        List<Driver> allDrivers = driverService.getAll();
+        allDrivers.forEach(System.out::println);
+
+        Car carRav4 = new Car("RAV4", toyotaManufacturer, List.of(driverIvan));
```
DV7Art, line 40: why u need List.of if u have allDrivers

DV7Art (approved):
Awesome

src/main/java/mate/jdbc/Main.java
ViktoriiaSemenchenko: Because in allDrivers I save all drivers and in carRav4 I use only one driver)

sarakhmen (approved):
Excellent job, keep it up :)

PR #644, opened 2023-06-14
Implemented CarService and CarDao classes

viprko:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,219 @@
...
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String query = "INSERT INTO cars (model, manufacturer_id) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            statement.setString(1, car.getModel());
+            statement.setLong(2, car.getManufacturer().getId());
```
ViktorBykh, line 28: Hereinafter, I think it's better to use setObject() for more flexibility:
```suggestion
            statement.setObject(2, car.getManufacturer().getId());
```

ViktorBykh (changes requested):
Everything looks great))
I have a few suggestions

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
viprko: I think I get what you mean, but that Is a statement and I set NUMBER to 2nds question mark. I presume that you mean to use getObject() when setting Long to model class

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,81 @@
...
     public static void main(String[] args) {
-        // test your code here
+        DatabaseInitializer databaseInitializer =
+                (DatabaseInitializer) injector.getInstance(DatabaseInitializer.class);
+        String[] initQueries = databaseInitializer.readFromFile(INIT_DB_FILE_PATH);
+        databaseInitializer.initializeDb(initQueries);
+
+        Manufacturer audiManufacturer = new Manufacturer("Audi", "Germany");
+        Manufacturer fiatManufacturer = new Manufacturer("Fiat", "Italy");
+        Manufacturer renaultManufacturer = new Manufacturer("Renault", "France");
```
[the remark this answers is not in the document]
viprko: I agree if it was tests classes, but here in the main we just try to make temporary tests

ViktorBykh (approved):
It looks good))

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,80 @@
...
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        manufacturerService.create(audiManufacturer);
+        manufacturerService.create(fiatManufacturer);
+        manufacturerService.create(renaultManufacturer);
+        System.out.println("get manufacturer with id = " + audiManufacturer.getId() + " --- "
+                + manufacturerService.get(audiManufacturer.getId()));
+        System.out.println("get all manufacturers --- " + manufacturerService.getAll());
+        Manufacturer toyotaManufacturer = new Manufacturer(2L, "Toyota", "Japan");
```
sarakhmen, line 27: why are you explicitly setting an id here? Also let's remove all hardcoded ids from your code and get it using getId() call of the corresponding model

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get cars by driver id: " + driverId);
+        }
+        if (!cars.isEmpty()) {
+            cars.forEach(car -> car.setDrivers(getDrivers(car.getId())));
+        }
```
sarakhmen, line 144:
```suggestion
        cars.forEach(car -> car.setDrivers(getDrivers(car.getId())));
```
This check is actually redundant

src/main/java/mate/jdbc/service/impl/CarsServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        Car carFromDb = get(car.getId());
+        if (!carFromDb.getDrivers().contains(driver)) {
+            carFromDb.getDrivers().add(driver);
+            update(carFromDb);
+        } else {
+            throw new RuntimeException(driver + "already in drivers list for " + car);
+        }
```
sarakhmen, line 50: As agreed with other mentors, we'd recommend removing this check and not throwing any exceptions here

src/main/java/mate/jdbc/service/impl/CarsServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        Car carFromDb = get(car.getId());
+        if (carFromDb.getDrivers().remove(driver)) {
+            update(carFromDb);
+        } else {
+            throw new RuntimeException(driver + " is not in the drivers list for " + car);
```
sarakhmen, line 59: the same here

sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

PR #647, opened 2023-06-14
build 0.9 std rev

maksigors:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,228 @@
...
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                car.setId(resultSet.getObject(1, Long.class));
+            }
+            return car;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create "
+                    + car + " !", e);
+        }
+    }
```
marimarig, line 40: I think you forgot to call the method addDriversToCar, before return car

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers().contains(driver)) {
+            throw new DataProcessingException("Driver "
+                    + driver.getName()
+                    + " has been already added to car "
+                    + car.getModel());
+        }
```
marimarig, line 56: You throw these exceptions in CarDaoImpl in update method. I don't think that need to duplicate it.

marimarig (approved):
Good job) but a few nuances

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
sarakhmen: agree with @marimarig , you should also add drivers to the created car

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
sarakhmen: As agreed with other mentors, we'd recommend removing this check and not throwing any exceptions here

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers().remove(driver)) {
+            carDao.update(car);
+        } else {
+            throw new DataProcessingException("Can't delete Driver "
+                    + driver.getName()
+                    + " from Car!"
+                    + car.getModel());
+        }
```
sarakhmen, line 70: the same here

sarakhmen (changes requested):
Let's improve your solution ;)

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
maksigors: yep, I've passed it

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
maksigors: no, it's two different methods addDriverToCar(Driver driver, Car car) & addDriversToCar(Car car)

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
maksigors: > 

the same)) removeDriver_FromCar(Driver driver, Car car) vs deleteDriver_s_FromCar(long carId)

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
sarakhmen: I'd say that it seems to be an overindulgence of exception usage. That's not critical whether already assigned to the car on addition or not on removal.

sarakhmen (approved):
Good job, you can improve your solution if you want :)

PR #648, opened 2023-06-14
Added homework solution

pkobevko:


ViktoriiaSemenchenko (approved):
good job!

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,60 @@
...
+            (CarService) injector.getInstance(CarService.class);
+    private static final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private static final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
     public static void main(String[] args) {
-        // test your code here
+        System.out.println("\n-----------------\nCheck creating new car");
+        Manufacturer manufacturer = new Manufacturer(0L, "ZAZ", "Ukraine");
```
maxline, line 23: We should not put `id` to the constructor. `id` will be generated by DB and returned in result of  `manufacturerService.create()` method.
Create constructor without `id` field.
Fix the same suggestion for other model classes.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT c.id AS car_id, model, mf.id AS manufacturer_id, mf.name AS "
+                + "manufacturer_name, mf.country AS manufacturer_country "
+                + "FROM cars AS c INNER JOIN manufacturers AS mf ON c.manufacturer_id = mf.id "
```
maxline, line 46: Let's start write `JOIN` from the new line for better readability. You can also remove `INNER` keyword because it is a default join type.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,212 @@
...
+            car.setDrivers(getDriversForCar(id));
+        }
+        return Optional.ofNullable(car);
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String getAllCarsQuery = "SELECT c.id AS car_id, model, mf.id AS manufacturer_id, "
+                + "mf.name AS manufacturer_name, mf.country AS manufacturer_country "
+                + "FROM cars AS c INNER JOIN manufacturers AS mf ON c.manufacturer_id = mf.id "
```
maxline, line 69: The same suggestion as in previous comment. Please check all queries and fix if needed.

maxline (changes requested):
Very good job! I left a few minor suggestions, take a look please.

PR #650, opened 2023-06-14
CarDao and CarService were implemented

m-pnsk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,124 @@
...
+        System.out.println(manufacturerService.delete(fiatManufacturer.getId()));
+        manufacturersMap.remove(fiatManufacturer.getName());
+
+        System.out.println("Get all:");
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        manufacturers.forEach(System.out::println);
+        return manufacturersMap;
+    }
+
+    private static Map<String, Driver> runExampleOfUsageDriverService(DriverService driverService) {
```
AnastasiiaVariuchuk, line 58: it's good idea to use a separate method, but maybe you should use simpler name

AnastasiiaVariuchuk: ok)

AnastasiiaVariuchuk (approved):

src/main/java/mate/jdbc/Main.java
sarakhmen: agree with @AnastasiiaVariuchuk let's make your methods names more concise and clear

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
```
sarakhmen, line 121: didn't you forget to assign also drivers to each found car?

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,218 @@
...
+        String request = "SELECT dc.id as car_id, m.name, dc.model, m.country, "
+                + "m.id as manufacturer_id "
+                + "FROM (SELECT * FROM cars "
+                + "WHERE id IN (SELECT car_id "
+                + "FROM cars_drivers "
+                + "WHERE is_deleted = FALSE AND driver_id = ?) "
+                + "AND is_deleted = FALSE"
```
sarakhmen, line 128: can you replace nested selects here with joins?

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers().contains(driver)) {
+            throw new RuntimeException("Driver already exist: " + driver);
+        }
```
sarakhmen, line 47: As agreed with other mentors, we'd recommend removing this check and not throwing any exceptions here

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        carDao.update(car);
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        if (car.getDrivers().remove(driver)) {
+            carDao.update(car);
+        } else {
+            throw new RuntimeException("Driver " + driver + " isn't the driver of car "
+                    + car + ". Deletion is not possible)");
```
sarakhmen, line 58: As agreed with other mentors, we'd recommend removing this check and not throwing any exceptions here

sarakhmen (changes requested):
let's improve your solution :)

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
m-pnsk: Yes, but this query is more optimized. JOIN is a more complicated operation than WHERE IN

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,221 @@
...
+                        = connection.prepareStatement(request)) {
+            getAllByDriverStatement.setLong(1, driverId);
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+            for (Car car : cars) {
+                car.setDrivers(getDriversByCar(car.getId()));
+            }
```
sarakhmen, line 143: nested connections issue. Please refer to common mistakes

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,221 @@
...
+            return cars;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't find all cars by driver id: "
+                    + driverId, e);
+        }
+    }
+
+    private void insertDrivers(Car car) {
+        String query = "INSERT INTO cars_drivers (car_id, driver_id) VALUES(?, ?);";
+        Long driverId = null;
```
sarakhmen, line 153: this driverId variable is redundant and could be removed

sarakhmen (changes requested):
Let's do some changes once again :)

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
sarakhmen: I think the purpose of this homework is to get used with joins. Also, it's quite hard to say 'where in' is more optimized than inner join. In most cases their performance is pretty equal. In some cases join is faster, in othes - where in

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
sarakhmen: oh, ok, didn't notice that :)

sarakhmen (approved):
Good job :)

PR #652, opened 2023-06-15
solution

volodymyrruso:


src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,75 @@
...
+    public List<Driver> getDrivers() {
+        return drivers;
+    }
+
+    public void setDrivers(List<Driver> drivers) {
+        this.drivers = drivers;
+    }
+
+    @Override
+    public String toString() {
```
nechupara, line 63: I would recommend you to generate `equals` and `hashCode` method just in case )

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+            getByIdStatement.setLong(1, id);
+            ResultSet resultSet = getByIdStatement.executeQuery();
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car by id " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriversForCar(id));
```
nechupara, line 64: I'm not sure, but maybe you can move `car.setDrivers(getDriversForCar(id))` inside the method `getCar(resultSet)`, because if you've initiated creating a car object then it means that car is not `null`.
But it's up to you.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+                PreparedStatement statement = connection.prepareStatement(getAll)) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                cars.add(getCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of cars from DB.", e);
+        }
+        for (Car car : cars) {
+            car.setDrivers(getDriversForCar(car.getId()));
```
nechupara, line 88: The same. I think you can move `car.setDrivers(getDriversForCar(id))` inside the method `getCar(resultSet)`, and get rid from a redundant second loop.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteQuery =
+                "UPDATE cars SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteStatement =
+                        connection.prepareStatement(deleteQuery)) {
+            deleteStatement.setLong(1, id);
+            return deleteStatement.executeUpdate() > 0;
```
nechupara, line 122: I would consider removing all car's drivers also, by calling `removeDriversFromCar(id)`

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+                + "WHERE cd.driver_id = ? AND c.is_deleted = FALSE;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllByDriverStatement =
+                        connection.prepareStatement(getAllByDriverRequest)) {
+            getAllByDriverStatement.setLong(1, driverId);
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            List<Car> cars = new ArrayList<>();
+            while (resultSet.next()) {
+                Car car = getCar(resultSet);
+                cars.add(car);
```
nechupara, line 146: I would suggest you to simplify it a little bit
```suggestion                
                cars.add(getCar(resultSet));
```

nechupara (changes requested):
Great job! Let's just consider some minor suggestions

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
nechupara: OK, I was wrong )

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
nechupara: and here I also was wrong ))

nechupara (approved):
nice )

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String insertFormatQuery =
```
sarakhmen, line 23:
```suggestion
        String insertQuery =
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+        removeDriversFromCar(car.getId());
+        insertDriver(car);
+        return car;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteQuery =
+                "UPDATE cars SET is_deleted = true WHERE id = ?;";
```
sarakhmen, line 117:
```suggestion
                "UPDATE cars SET is_deleted = TRUE WHERE id = ?;";
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteStatement =
+                        connection.prepareStatement(deleteQuery)) {
+            deleteStatement.setLong(1, id);
+            removeDriversFromCar(id);
```
sarakhmen, line 122: pay attention to the nested connections problem, which is described in common mistakes. I'd recommend adding 'on delete cascade' to the corresponding table instead

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,223 @@
...
+            return deleteStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete car with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        String getAllByDriverRequest = "SELECT c.id AS car_id, c.model, c.manufacturer_id, "
```
sarakhmen, line 131: variable naming issues. Let's use 'query' suffix instead of 'request' as it was done in your previous methods

sarakhmen (changes requested):
Let's improve your solution :)

rovein (approved):
Great job!

PR #658, opened 2023-06-18
Join work

Hlib13:


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        String query = "INSERT INTO cars_drivers (driver_id, car_id) VALUES (?, ?)";
+        if (!isDriverRelateCar(driver.getId(), car.getId())) {
+            try (Connection connection = ConnectionUtil.getConnection();
```
okuzan, line 53: image

okuzan (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Optional;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
```
olekskov, line 19: I can't see any logic in dao layer that will put or get information regarding drivers.
For the moment it seems that you simply ignore them on this layer

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,61 @@
...
+        drivers.add(john);
+
+        Driver alice = new Driver();
+        alice.setName("Alice");
+        drivers.add(alice);
+
+        car.setDrivers(drivers);
+
+        List<Car> allCarsUpdate = carService.getAll();
+        System.out.println("All cars: " + allCarsUpdate);
```
olekskov, line 59: You set up list of drivers, assign it to car and then just use getAll() totally ignoring car and list of drivers you created.
What is the purpose of such code?

olekskov (changes requested):

aaananas (approved):

PR #659, opened 2023-06-19
Completed the task, the program works correctly

VRuban373:


Fafakepa (approved):
Overall, good job, but i can`t see CarDaoImpl, maybe it`s bug)

maxline (changes requested):
Very good job! I left just a few minor comments.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,64 @@
...
+    private static final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+    private static final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+
     public static void main(String[] args) {
-        // test your code here
+        Manufacturer zaz = manufacturerService.create(new Manufacturer("Zaz", "Ukraine"));
+        Manufacturer dacia = manufacturerService.create(new Manufacturer("Dacia", "Romania"));
+        Manufacturer seat = manufacturerService.create(new Manufacturer("Seat", "Spain"));
```
maxline, line 25: `Manufacturer` variables also can be final because its never change in the code below.

maxline (approved):
Well done! Approved

PR #660, opened 2023-06-20
Create class Car, CarServcice interface with implementation

VolodymyrDavidov:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,52 @@
...
+        carUpdate.setDrivers(listUp);
+
+        System.out.println(carService.create(car));
+        System.out.println(carService.get(21L));
+        carService.getAll().forEach(System.out::println);
+        System.out.println(carService.update(carUpdate));
+        System.out.println(carService.delete(2L));
+        carService.addDriverToCar(driverTaxi, carUpdate);
+        carService.removeDriverFromCar(driverTaxi, carUpdate);
+        carService.getAllByDriver(1L).forEach(System.out::println);
```
okuzan, line 44: Ids should be handled by databases. Don't hardcode them where you can avoid this. Retrieve ids from existing objects using getters.

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,73 @@
...
+        return Objects.equals(getId(), car.getId())
+                && Objects.equals(getModel(), car.getModel())
+                && Objects.equals(getManufacturer(), car.getManufacturer())
+                && Objects.equals(getDrivers(), car.getDrivers());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, model, manufacturer, drivers);
+    }
```
okuzan, line 62: not necessary in this task

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,201 @@
...
+                drivers.add(getDriver(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get drivers "
+                    + "from cars_drivers by id " + carId, e);
+        }
+        return drivers;
+    }
+
+    private void deleteDriversFromCar(Long carsId) {
```
okuzan, line 174:
```suggestion
    private void deleteDriversFromCar(Long carId) {
```

PR #663, opened 2023-06-30
carService implementation

krismva:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,48 @@
...
+        Driver nik = new Driver("Nik", "332233");
+        Driver mihael = new Driver("Mihael", "887788");
+
+        driverService.create(john);
+        driverService.create(andrew);
+        driverService.create(nik);
+        driverService.create(mihael);
+        System.out.println(driverService.getAll());
+
+        Car car = new Car("nissan", manufacturer, List.of(john, andrew));
```
olekskov, line 36: You are creating a Car object with model 'nissan' and immediately after you are changing the model to 'mazda'.
Probably you wanted to test update operation here?

PR #665, opened 2023-07-04
Work branch

RostyslavOnysh:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+    private static void insertCarDrivers(Car car) {
+        String query = "INSERT INTO cars_drivers (car_id, driver_id) "
+                + "VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            for (Driver driver : car.getDrivers()) {
+                statement.setLong(1, car.getId());
+                statement.setLong(2, driver.getId());
+                statement.addBatch();
+                statement.clearParameters();
```
kozhukhovsky, line 218:
```suggestion
                statement.addBatch();
                statement.clearParameters();
```

why do u need this?

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,226 @@
...
+            }
+            statement.executeBatch();
```
kozhukhovsky, line 220:
```suggestion
            statement.executeUpdate();
```

kozhukhovsky (approved):

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
RostyslavOnysh: It seems like I got a bit too caught up with it and confused myself.... 🙈  To avoid executing each query individually, I add them to a batch using addBatch(). Then, I use executeBatch() to execute the entire batch as a single query to the database. I just realized that I don't need to call clearParameters() and addBatch() inside the loop. I only need to call addBatch() once inside the loop for each query, and then use executeBatch() outside the loop to execute the entire batch.

PR #666, opened 2023-07-07
impemented car database

vaniii1:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,53 @@
...
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer1 = new Manufacturer();
```
Ivan95kos, line 19: it is bad practice to use numbers in names

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,53 @@
...
+import mate.jdbc.service.CarService;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService =
```
Ivan95kos, line 17: All services should be class fields

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,207 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't insert book into DB", e);
+        }
+        createDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Car get(Long id) {
```
Ivan95kos, line 43: This method should return `Optional`
```suggestion
    public Optional<Car> get(Long id) {
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id);
```
Ivan95kos, line 24: This method should be like this:
```
    @Override
    public Car get(Long id) {
        return carDao.get(id)
                .orElseThrow(() -> new NoSuchElementException("Could not get car by id = " + id));
    }
```

Ivan95kos (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,56 @@
...
+        driverService.create(secondDriver);
+        driverService.create(thirdDriver);
+        Car car = new Car();
+        List<Driver> drivers = new ArrayList<>();
+        drivers.add(firstDriver);
+        drivers.add(secondDriver);
+        car.setDrivers(drivers);
+        car.setManufacturer(manufacturer);
+        car.setModel("vitara");
+        CarService carService = (CarService) injector.getInstance(CarService.class);
```
Ivan95kos, line 46: You have this field as class field
```suggestion
```

Ivan95kos (approved):

PR #667, opened 2023-07-07
jv-join. Pochtalon

pochtalon:
Created class Car;
Created CarService and CarDao interfaces;
Created CarServiceImpl and CarDaoImpl classes;
Created table cars and added changes to init_db.sql file; Created table cars_drivers and added changes to init_db.sql file; Implemented methods in CarServiceImpl class;
Implemented methods in CarDaoImpl class;

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
@@ -1,0 +1,209 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import mate.jdbc.dao.CarDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,209 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create car "
+                    + car + ". ", e);
+        }
+        insertDrivers(car);
+        return car;
+    }
+
+    @Override
+    public Car get(Long id) {
```
Ivan95kos, line 43: This method should return `Optional`
```suggestion
    public Optional<Car> get(Long id) {
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    @Inject
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
```
Ivan95kos, line 23: This method should be like this:
```
    @Override
    public Car get(Long id) {
        return carDao.get(id)
                .orElseThrow(() -> new NoSuchElementException("Could not get car by id = " + id));
    }
```

Ivan95kos (changes requested):
Good job overall!

Ivan95kos (approved):
Nice

PR #668, opened 2023-07-10
created the Car class, implemented methods in the CarDaoImpl and CarS…

0-zagar-0:
…erviceImpl classes and tested all methods in the Main class

### DIFF
--- a/src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
@@ -51,11 +51,18 @@ public class CarServiceImpl implements CarService {
             car.setDrivers(drivers);
         }
         drivers.add(driver);
+        
+        if (carDao.get(car.getId()).isPresent()) {
+            carDao.update(car);
+        }
     }
     
     @Override
     public void removeDriverFromCar(Driver driver, Car car) {
-        car.getDrivers().remove(car);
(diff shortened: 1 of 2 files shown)


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        
+        if (drivers == null) {
+            drivers = new ArrayList<>();
+            car.setDrivers(drivers);
+        }
+        drivers.add(driver);
```
olekskov, line 53: You changes list of drivers but do not save changes to DB

olekskov (changes requested):

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+        return carDao.delete(id);
+    }
+    
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        
+        if (drivers == null) {
+            drivers = new ArrayList<>();
+            car.setDrivers(drivers);
```
okuzan, line 51:
```suggestion
            car.setDrivers(new ArrayList<>());
```

okuzan (approved):
Well done!

PR #670, opened 2023-07-11
solution

VictoriaMrkvch:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,20 @@
...
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Car car = carService.get(3L);
+        Driver driver = driverService.get(2L);
+        carService.addDriverToCar(driver, car);
+        carService.getAll().forEach(System.out::println);
```
okuzan, line 18: It's not enough, let's create here drivers, manufacturers, cars (using created drivers). Try to remove driver from car too. Don't access them by hardcoded IDs, use getId() after saving in DB

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
okuzan, line 9: Dao should return optional for get, fix implementation accordingly

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+                PreparedStatement statement = connection.prepareStatement(selectRequesr)) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                car = getCar(resultSet);
+                cars.add(car);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get cars from DB", e);
+        }
+        for (Car c: cars) {
```
okuzan, line 84:
```suggestion
        for (Car car : cars) {
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get car by id " + id, e);
+        }
+        if (car != null) {
+            car.setDrivers(getDriversForCar(id));
+        }
+        return car;
```
okuzan, line 62: use Optional.ofNullable

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id);
```
okuzan, line 23: here use orElseThrow on optional returned by dao

okuzan (changes requested):
Good job overall!

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                PreparedStatement statement = connection.prepareStatement(selectRequesr)) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                car = getCar(resultSet);
+                cars.add(car);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get cars from DB", e);
+        }
+        for (Car car1: cars) {
```
kozhukhovsky, line 86:
```suggestion
        for (Car currentCar: cars) {
```

It's better to name something like this, but do not use c or a name with a number

kozhukhovsky (approved):

PR #675, opened 2023-07-18
hw-jv-join-solution-andrey-kiyas

andrey-kiyas:


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.CarService;
+
+@Service
+public class CarServiceImpl implements CarService {
+    @Inject
+    private final CarDao carDao = new CarDaoImpl();
```
Rommelua, line 16: @Inject should work
```suggestion
    private final CarDao carDao;
```

Rommelua (approved):

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
andrey-kiyas: Yes, I overdid it. Thank you!

PR #677, opened 2023-07-21
Car, СarService and CarDao are implemented

YaroslavKostenko89:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,55 @@
...
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        CarService carService = (CarService) injector.getInstance(CarService.class);
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Car car = carService.get(2L);
```
Ivan95kos, line 18: don't hardcode ids, retrieve by getId()

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+                for (Driver driver : drivers) {
+                    if (driver.getId().equals(driverId)) {
+                        result.add(car);
+                        break;
+                    }
+                }
+            }
+        }
+        return result;
+    }
```
Ivan95kos, line 76: It is not a good way to search,
Better to create a SQL query in DB
```suggestion
    @Override
    public List<Car> getAllByDriver(Long driverId) {
        return carDao.getAllByDriver(driverId);
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,59 @@
...
-        // test your code here
+        Car car = new Car();
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer(1L, "Nisan", "Japan");
+        car.setManufacturer(manufacturerService.create(manufacturer));
+        car.setModel("X1");
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        List<Driver> drivers = new ArrayList<>();
+        Driver driverBob = new Driver(1L, "Bob", "qw5890");
```
Ivan95kos, line 25: You should not set the ID
The DB creates IDs `AUTO_INCREMENT`
```suggestion
        Driver driverBob = new Driver("Bob", "qw5890");
```

Ivan95kos (approved):

PR #678, opened 2023-07-21
Hw join

AntonU91:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService =
+                (ManufacturerService) INJECTOR.getInstance(ManufacturerService.class);
+
+        Car honda = new Car();
+        honda.setModel("RCV");
+
+        honda.setManufacturer(manufacturerService.get(26L));
```
Ivan95kos, line 22: don't hardcode ids, retrieve by getId()

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        Driver driver = driverService.get(driverId);
+        List<Car> cars = carDao.getAll();
+        return cars.stream()
+                   .filter(car -> car.getDrivers().contains(driver))
+                   .collect(Collectors.toList());
+    }
```
Ivan95kos, line 67: It is not a good way to search,
Better to create a SQL query in DB
```suggestion
    @Override
    public List<Car> getAllByDriver(Long driverId) {
        return carDao.getAllByDriver(driverId);
    }
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
AntonU91: Done

src/main/java/mate/jdbc/Main.java
AntonU91: Done

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,230 @@
...
+                statement.setLong(1, driver.getId());
+            }
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException(
+                    "Could not insert drivers for car with id " + car.getId(), e);
+        }
+    }
+
+    public List<Car> getAllByDriver(Long driverId) {
```
olekskov, line 198: All public method should be placed before all private ones

olekskov (approved):
Approved, but please check comment below for further tasks

olekskov (approved):

PR #679, opened 2023-07-23
Hometask jv-join completed

MarynaKushniruk:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,225 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(queryGetAllCars);) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                carList.add(getCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get a list of cars", e);
+        }
+        if (!carList.isEmpty()) {
```
olekskov, line 81: It's not necessary to check if the list is empty before the for-each loop. If the list is empty, the loop will not be executed.

olekskov (approved):
Good job

PR #680, opened 2023-07-23
added solution

Serhii-Nosulia:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,229 @@
...
+            for (Car car : cars) {
+                car.setDrivers(getDriversList(car.getId()));
+            }
+        }
+        return cars;
+    }
+
+    private Car parseCar(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setId(resultSet.getObject(1, Long.class));
```
olekskov, line 163: While setting the id of the manufacturer, you are using getObject(1, Long.class) which will retrieve the car's id instead of the manufacturer's id. You should use resultSet.getObject('manufacturer_id', Long.class) instead to get the correct id.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,63 @@
...
+        Driver john = drSer.get(6L);
+        Driver alice = drSer.get(7L);
+
+        List<Driver> drivers = new ArrayList<>();
+        drivers.add(bob);
+        drivers.add(john);
+        drivers.add(alice);
+
+        Driver ivan = new Driver();
+        ivan.setName("Ivan");
```
olekskov, line 34: You are directly setting values to the 'Driver' object. It's better to use a constructor or a builder pattern to initialize an object.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,63 @@
...
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+    private static final ManufacturerService manSer =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private static final DriverService drSer =
+            (DriverService) injector.getInstance(DriverService.class);
+    private static final CarService carSer =
```
olekskov, line 19: abbreviations mostly never provide benefit but usually lead to problems with code reading
```suggestion
    private static final ManufacturerService manufacturerService =
            (ManufacturerService) injector.getInstance(ManufacturerService.class);
    private static final DriverService driverService =
            (DriverService) injector.getInstance(DriverService.class);
    private static final CarService carSer =
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,229 @@
...
+                cars.add(parseCar(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get a list of cars from cars table", e);
+        }
+        if (!cars.isEmpty()) {
+            for (Car car : cars) {
+                car.setDrivers(getDriversList(car.getId()));
+            }
+        }
```
olekskov, line 93: If cars is empty you will just have zero iterations of cycle:
```suggestion
        for (Car car : cars) {
            car.setDrivers(getDriversList(car.getId()));
        }
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,229 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get all cars by driver with id: "
+                    + driverId, e);
+        }
+        if (!cars.isEmpty()) {
+            for (Car car : cars) {
+                car.setDrivers(getDriversList(car.getId()));
+            }
+        }
```
olekskov, line 157: same

olekskov (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,57 @@
...
+            (DriverService) injector.getInstance(DriverService.class);
+    private static final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+
     public static void main(String[] args) {
-        // test your code here
+
+        Driver bob = driverService.get(5L);
+        Driver john = driverService.get(6L);
+        Driver alice = driverService.get(7L);
```
okuzan, line 26: you shouldn't rely on hardcoded ids even in main

okuzan (approved):
Well done!

PR #681, opened 2023-07-25
add my solution

mykhailoKrlKrk:


src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,22 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+
+public interface CarDao {
+
```
okuzan, line 8:
```suggestion
```

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,22 @@
...
+    Car create(Car car);
+
+    Car get(Long id);
```
okuzan, line 11: get on Dao should always return optional - dao's job doesn't include business logic of how to handle the case where required entity is missing, we are returning optional to the Service layer, and only there we will deal with this

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+
```
okuzan, line 20:
```suggestion
```

src/main/java/mate/jdbc/service/CarService.java
```diff
@@ -0,0 +1,29 @@
...
+
+    void addDriverToCar(Driver driver, Car car);
+
+    void removeDriverFromCar(Driver driver, Car car);
+
+    List<Car> getAllByDriver(Long driverId);
+
+    List<Driver> getAllDriversByCar(Long carId);
+
+    Car createDefaultCar(Manufacturer manufacturer, Driver driver);
```
okuzan, line 28: If you only need it for testing it probably shouldn't be in public API, just have this method in Main

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers() != null) {
+            car.getDrivers().add(driver);
+        }
+    }
+
+    @Override
+    public void removeDriverFromCar(Driver driver, Car car) {
+        car.getDrivers().remove(driver);
```
okuzan, line 52: don't forget to update the car in db to persist the changes

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        if (car.getDrivers() != null) {
+            car.getDrivers().add(driver);
```
okuzan, line 46: same here changes won't persist

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                cars.add(getCar(resultSet));
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private Car parseCarWithManufacturerResultSet(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
```
okuzan, line 155: private methods should go after public ones

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteCarQuery = "UPDATE cars SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement softDeleteCarStatement =
+                        connection.prepareStatement(deleteCarQuery)) {
+            softDeleteCarStatement.setLong(1,id);
```
okuzan, line 127: autoformat the file
```suggestion
            softDeleteCarStatement.setLong(1, id);
```

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return car;
+    }
+
+    @Override
+    public Car get(Long id) {
+        String selectRequest = "SELECT c.id AS car_id,model,manufacturer_id,m.name,country "
+                + "FROM CARS c "
+                + "JOIN MANUFACTURERS m "
+                + "ON c.manufacturer_id = m.id WHERE c.id = ? AND c.is_deleted = false;";
```
okuzan, line 49: place where on the new line

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        if (car != null) {
+            car.setDrivers(getDriversForCar(id));
+        }
+        return car;
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String getAllRequest = "SELECT * FROM cars WHERE is_deleted = false;";
```
okuzan, line 71: you should join here same as in get method

src/main/resources/init_db.sql
```diff
@@ -18,3 +18,23 @@ CREATE TABLE `drivers` (
...
+                        `manufacturer_id` bigint DEFAULT NULL,
+                        PRIMARY KEY (`id`),
+                        KEY `cars_manufacturer_fk` (`manufacturer_id`),
+                        CONSTRAINT `cars_manufacturer_fk` FOREIGN KEY (`manufacturer_id`) REFERENCES `manufacturers` (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=57 DEFAULT CHARSET=utf8mb3;
+
+CREATE TABLE `cars_drivers` (
+                                `car_id` bigint NOT NULL,
+                                `driver_id` bigint NOT NULL,
+                                `is_deleted` tinyint NOT NULL DEFAULT '0',
```
okuzan, line 35: cars_drivers shouldn't implement soft delete technique

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            }
+            return cars;
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Car update(Car car) {
+        String updateQuery = "UPDATE cars SET model = ?, manufacturer_id = ? "
```
okuzan, line 88: let's structure this method as follows, update model and  manufacturer_id, then after the connection is closed, delete and insert drivers (use dedicated private methods)

okuzan (changes requested):

src/main/resources/init_db.sql
mykhailoKrlKrk: Ok,how i can delete all relations for some car without soft delete?

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateFieldsStatement =
+                        connection.prepareStatement(updateQuery);) {
+            updateFieldsStatement.setString(1, car.getModel());
+            updateFieldsStatement.setLong(2, car.getManufacturer().getId());
+            updateFieldsStatement.setLong(3, car.getId());
+            updateFieldsStatement.executeUpdate();
+
+            delete(car.getId());
+            insertDrivers(car);
```
kozhukhovsky, line 105:
```suggestion
            delete(car.getId());
            insertDrivers(car);
```

after try/catch block

src/main/resources/init_db.sql
kozhukhovsky: First you do in cara is_deleted true, and then for all the data on this car from cars_driver you do exactly delete.
Just an additional method and a separate request, as in other places.

kozhukhovsky (changes requested):
Need small fixes

okuzan (approved):
Good job!

PR #682, opened 2023-07-25
jv-join

s-hrynus:
…ssary methods and tested they in main

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
@@ -1,0 +1,195 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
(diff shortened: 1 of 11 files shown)


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,195 @@
...
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+
```
okuzan, line 21: line

okuzan (approved):
Well done!

PR #684, opened 2023-07-26
added the Car class to the model,

TemkaHillfinger:
added the CarService interface and CarDao interface added the CarServiceImpl and CarDaoImpl and implemented the corresponding interfaces there

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -18,7 +18,7 @@ import mate.jdbc.util.ConnectionUtil;
 public class DriverDaoImpl implements DriverDao {
     @Override
     public Driver create(Driver driver) {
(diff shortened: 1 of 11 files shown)


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -18,7 +18,7 @@
 public class DriverDaoImpl implements DriverDao {
     @Override
     public Driver create(Driver driver) {
-        String query = "INSERT INTO drivers (name, license_number) VALUES (?, ?);";
+        String query = "INSERT INTO driver (name, license_number) VALUES (?, ?);";
```
Ivan95kos, line 21: The table must be named in the plural
```suggestion
        String query = "INSERT INTO drivers (name, license_number) VALUES (?, ?);";
```

Ivan95kos (approved):

PR #685, opened 2023-07-26
Add working solution: connected to MySQL, add tables to DB: cars, car…

WasjaZeikan:
…s_drivers, created model Car, its DAO, CarService and its implementation, tested CarService in method main

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
@@ -18,12 +18,43 @@ import mate.jdbc.util.ConnectionUtil;
 
 @Dao
 public class CarDaoImpl implements CarDao {
+    private static final String CREATE_CAR_REQUEST = "INSERT INTO "
+            + "cars(manufacturer_id, model) VALUES(?, ?);";
+    private static final String GET_CAR_BY_ID_REQUEST = "SELECT c.id AS car_id,"
+            + " c.model, m.id AS manufacturer_id,"
+            + " m.name, m.country FROM cars c INNER JOIN manufacturers m "
+            + "ON m.id = c.manufacturer_id WHERE c.is_deleted = FALSE "
+            + "AND m.is_deleted = FALSE;";
+    private static final String GET_ALL_CARS_REQUEST = "SELECT c.id AS car_id, "
+            + "c.model, m.id AS manufacturer_id,"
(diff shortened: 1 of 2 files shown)


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,202 @@
...
+import mate.jdbc.model.Car;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class CarDaoImpl implements CarDao {
+    @Override
+    public Car create(Car car) {
+        String createRequest = "INSERT INTO cars(manufacturer_id, model) VALUES(?, ?);";
```
sokimaaa, line 23: You can extract all query-string into constants for the sake of readability.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,202 @@
...
+            createStatement.executeUpdate();
+            ResultSet generatedKeys = createStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                car.setId(generatedKeys.getObject(1, Long.class));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert car: " + car
+                    + " to DB", e);
+        }
+        insertDrivers(car);
```
sokimaaa, line 38: Do not implement following methods in DAO layer, only in service layer. Use update method from DAO.

void addDriverToCar(Driver driver, Car car);
void removeDriverFromCar(Driver driver, Car car);

see https://mate-academy.github.io/jv-program-common-mistakes/java-JDBC/join/Joins_checklist.html

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    private CarDao carDao;
+
+    @Override
+    public Car create(Car car) {
+        return carDao.create(car);
+    }
+
+    @Override
+    public Car get(Long id) {
+        return carDao.get(id).orElseThrow();
```
sokimaaa, line 23: let's throw a dedicated exception with the explanation.

sokimaaa (changes requested):

Rommelua (approved):

PR #686, opened 2023-07-26
Solution of this task

IvanovYurii:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+        addDriversToCar(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String selectCarByIdRequest = "SELECT c.id as car_id, model, c.manufacturer_id,"
+                + " m.name, m.country FROM cars c JOIN manufacturers m "
+                + "ON c.manufacturer_id = m.id WHERE c.id = ? AND c.is_deleted = 0;";
```
kozhukhovsky, line 46: An error from a previous issue. For boolean it's better to use True/False

for all queries

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,246 @@
...
+                        }
+                        car.getDrivers().add(driver);
+                        break;
+                    }
+                }
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get drivers for cars from cars DB.", e);
+        }
+        return cars;
```
kozhukhovsky, line 104:
```suggestion
        String selectAllCarAddDriversRequest = "SELECT cd.cars_id, d.id, d.name, "
                + "d.license_number FROM cars_drivers cd "
                + "JOIN drivers d ON cd.drivers_id = d.id;";
        try (Connection connection = ConnectionUtil.getConnection();
                PreparedStatement driversStatement = connection
                        .prepareStatement(selectAllCarAddDriversRequest)) {
            ResultSet resultSet = driversStatement.executeQuery();
            while (resultSet.next()) {
                Long carId = resultSet.getLong("cars_id");
                Driver driver = getDriver(resultSet);
                for (Car car : cars) {
                    if (car.getId().equals(carId)) {
                        if (car.getDrivers() == null) {
                            car.setDrivers(new ArrayList<>());
                        }
                        car.getDrivers().add(driver);
                        break;
                    }
                }
            }
        } catch (SQLException e) {
            throw new DataProcessingException("Couldn't get drivers for cars from cars DB.", e);
        }
        return cars;
```

move this to the separate method

kozhukhovsky (changes requested):

kozhukhovsky (approved):

PR #687, opened 2023-07-30
completed task

BakhmetIvan:


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+        car.setDrivers(drivers);
+        carDao.update(car);
+    }
+
+    @Override
+    public List<Car> getAllByDriver(Long driverId) {
+        return null;
+    }
```
Ivan95kos, line 68: Why does the method return the null?
```suggestion
    @Override
    public List<Car> getAllByDriver(Long driverId) {
        return carDao.getAllByDriver(driverId);
    }
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        List<Driver> drivers = car.getDrivers();
+        for (int i = 0; i < drivers.size(); i++) {
+            if (drivers.get(i).equals(driver)) {
+                drivers.remove(i);
+                break;
+            }
+        }
+        car.setDrivers(drivers);
+        carDao.update(car);
+    }
```
Ivan95kos, line 63: It is better to do like:
```suggestion
    @Override
    public void removeDriverFromCar(Driver driver, Car car) {
        car.getDrivers().remove(driver);
        carDao.update(car);
    }
```

src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
+    }
```
Ivan95kos, line 50: also
```suggestion
    @Override
    public void addDriverToCar(Driver driver, Car car) {
        car.getDrivers().add(driver);
        carDao.update(car);
    }
```

Ivan95kos (changes requested):

olekskov (approved):
good job

PR #690, opened 2023-08-03
jv-join

OleksiiMakovkin:
Created:
-Inreface CarDao
-Interface CarService
-Class Car
-Class CarDaoImpl
-Class CarServiceImpl

### DIFF
--- a/src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
@@ -1,0 +1,65 @@
+package mate.jdbc.service.impl;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import mate.jdbc.dao.CarDao;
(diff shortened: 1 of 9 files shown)


src/main/java/mate/jdbc/service/impl/CarServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    public boolean delete(Long id) {
+        return carDao.delete(id);
+    }
+
+    @Override
+    public void addDriverToCar(Driver driver, Car car) {
+        List<Driver> drivers = car.getDrivers();
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carDao.update(car);
```
aaananas, line 48:
```suggestion
        car.getDrivers().add(driver);
        carDao.update(car);
```

aaananas (approved):

PR #692, opened 2023-08-08
Homework jv-join.

UniyatYura:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        insertDriver(car);
+        return car;
+    }
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String selectCarRequest = "SELECT cars.id AS car_id, name, model, country, "
+                + "manufacturers.id AS manufacturer_id FROM cars JOIN manufacturers ON"
+                + " cars.manufacturer_id = manufacturers.id WHERE cars.id = ?"
+                + " AND cars.is_deleted = FALSE;";
```
olekskov, line 47: Please try to split your queries in more tidy and readable way. Spaces are better in the end of the row. Rows should be logically split to have full logical command in single row:
```suggestion
        String selectCarRequest = "SELECT cars.id AS car_id, name, model, country, "
                + "manufacturers.id AS manufacturer_id "
                + "FROM cars " 
                + "JOIN manufacturers ON cars.manufacturer_id = manufacturers.id "
                + "WHERE cars.id = ? "
                + "AND cars.is_deleted = FALSE;";
```

olekskov (approved):
Good job, approved.
Please check comment below for further tasks

PR #693, opened 2023-08-12
push(alfa): push task for finding solution for Injector.getInstance

urlo0k:


olekskov (changes requested):
Please fix build failed due to checkstyle before we may proceed with review

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,58 @@
...
+        driverService.create(driverAmina);
+
+        CarService carService =
+                (CarService) injector.getInstance(CarService.class);
+
+        Car carTeslaOne = new Car(null, "Tesla", manufacturerTesla, null);
+        carService.create(carTeslaOne);
+
+        Car carTeslaTwo = new Car(null, "Tesla", manufacturerTesla, null);
+        carService.create(carTeslaTwo);
```
okuzan, line 37: Consider creating a constructor without the 'id' parameter. In most cases, the 'id' will be generated automatically during persistence in the database, and it's uncommon to set it manually.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,58 @@
...
+        System.out.println(carTeslaOne);
+
+        List<Car> allCarsDriverArtem = carService.getAllByDriver(driverAmina.getId());
+        for (Car car : allCarsDriverArtem) {
+            System.out.println(car);
+        }
+
+        System.out.println(carService.get(55L));
+
+        carService.delete(55L);
```
okuzan, line 54: don't rely on hardcoded ids, use getId on returned entites

src/main/java/mate/jdbc/dao/CarDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Car;
+
+public interface CarDao {
+    Car create(Car car);
+
+    Car get(Long id);
```
okuzan, line 9: get in dao should return optional

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,212 @@
...
+                resultSet.getObject("car_id", Long.class),
+                resultSet.getString("model"),
+                new Manufacturer(
+                        resultSet.getObject("manufacturer_id", Long.class),
+                        resultSet.getString("name"),
+                        resultSet.getString("country")), null);
+    }
+
+    private List<Driver> getDriversFromCar(Long carId) {
+        String getAllDriversFromCarRequest = "SELECT id, name, lisense_number "
```
okuzan, line 190: not a mistake, but more idiomatic to use query in that context

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,212 @@
...
+                driver.setLicenseNumber(resultSet.getString("lisense_number"));
+                drivers.add(driver);
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't find drivers in DB by car carId " + carId, e);
+        }
+    }
+}
+
```
okuzan, line 212:
```suggestion
```

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,66 @@
...
+    private Long id;
+    private String model;
+    private Manufacturer manufacturer;
+    private List<Driver> drivers;
+
+    public Car() {
+    }
+
+    public Car(Long id, String model, Manufacturer manufacturer, List<Driver> drivers) {
+        this.id = id;
```
okuzan, line 15:
```suggestion
    public Car(String model, Manufacturer manufacturer, List<Driver> drivers) {
```

okuzan (changes requested):
liCense misspelled :)

src/main/java/mate/jdbc/dao/CarDao.java
urlo0k: I will do it. But by the conditions of the HW 04: ...
CarService methods:
...
- Car get(Long id);

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
urlo0k: Sorry, I can't understand what you mean by the definition of "more idiomatic".
Respectively, what I had to correct?

olekskov (approved):

PR #694, opened 2023-08-14
Solution

Volodymyr-Mykychak:


src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,208 @@
...
+        car.setDrivers(driversFromCar);
+        return Optional.ofNullable(car);
+    }
+
+    @Override
+    public List<Car> getAll() {
+        String query = "SELECT cars.id, cars.manufacturer_id, cars.model, "
+                + "manufacturers.id, manufacturers.name, manufacturers.country FROM cars "
+                + "INNER JOIN manufacturers "
+                + "ON cars.manufacturer_id = manufacturers.id AND cars.is_deleted = false;";
```
overpathz, line 70: image

Please, review your code on common mistakes from the checklist and request a review again. Thanks!

overpathz (changes requested):

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -6,10 +6,10 @@
 import java.util.Properties;
 
 public class ConnectionUtil {
-    private static final String URL = "YOUR DATABASE URL";
-    private static final String USERNAME = "YOUR USERNAME";
-    private static final String PASSWORD = "YOUR PASSWORD";
-    private static final String JDBC_DRIVER = "YOUR DRIVER";
+    private static final String URL = "jdbc:mysql://localhost:3306/taxi_service";
+    private static final String USERNAME = "user";
```
olekskov, line 10: Hardcoding credentials in the source code is a bad practice. It would be better to use configuration files or environment variables. This way, you can avoid accidents like committing sensitive data into a public repository.

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -6,10 +6,10 @@
 import java.util.Properties;
 
 public class ConnectionUtil {
-    private static final String URL = "YOUR DATABASE URL";
-    private static final String USERNAME = "YOUR USERNAME";
-    private static final String PASSWORD = "YOUR PASSWORD";
-    private static final String JDBC_DRIVER = "YOUR DRIVER";
+    private static final String URL = "jdbc:mysql://localhost:3306/taxi_service";
```
olekskov, line 9: Hardcoding credentials in the source code is a bad practice. It would be better to use configuration files or environment variables. This way, you can avoid accidents like committing sensitive data into a public repository.

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -6,10 +6,10 @@
...
+    private static final String USERNAME = "user";
+    private static final String PASSWORD = "password";
```
olekskov, line 11: Hardcoding credentials in the source code is a bad practice. It would be better to use configuration files or environment variables. This way, you can avoid accidents like committing sensitive data into a public repository.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            ResultSet resultSet = getStatment.executeQuery();
+            if (resultSet.next()) {
+                car = getCar(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get car by id " + id, e);
+        }
+        List<Driver> driversFromCar = getDriversFromCar(id);
+        car.setDrivers(driversFromCar);
+        return Optional.ofNullable(car);
```
olekskov, line 63: Instead of returning an empty Optional if the car is not found, it would be better to return Optional.empty(). This makes it clearer that no value is present.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @Override
+    public Optional<Car> get(Long id) {
+        String query = "SELECT cars.id, cars.manufacturer_id, cars.model, manufacturers.name, "
+                + "manufacturers.country "
+                + "FROM cars "
+                + "INNER JOIN manufacturers "
+                + "ON cars.manufacturer_id = manufacturers.id "
+                + "WHERE cars.id = ? AND cars.is_deleted = FALSE;";
+        Car car = new Car();
```
olekskov, line 50: It's not a good idea to create a new Car object before you check if it exists in the database. Instead, initialize the car object inside the if block after the resultSet.next() check.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        for (Long carID : carsID) {
+            carList.add(get(carID).get());
+        }
+        return carList;
+    }
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        car.setId(resultSet.getObject("cars.id", Long.class));
```
olekskov, line 155: The column alias should be 'cars.id' in the resultSet.getObject() method.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            ResultSet resultSet = getAllByDriverStatement.executeQuery();
+            while (resultSet.next()) {
+                Long carID = resultSet.getObject("cars_drivers.car_id", Long.class);
+                carsID.add(carID);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get cars by driver id " + driverId, e);
+        }
+        for (Long carID : carsID) {
+            carList.add(get(carID).get());
```
olekskov, line 148: Use isPresent() before get() to avoid NoSuchElementException.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return carList;
+    }
+
+    private Car getCar(ResultSet resultSet) throws SQLException {
+        Car car = new Car();
+        car.setId(resultSet.getObject("cars.id", Long.class));
+        car.setModel(resultSet.getString("cars.model"));
+        Manufacturer manufacturer = new Manufacturer(
+                resultSet.getObject("manufacturer_id", Long.class),
```
olekskov, line 158: The column alias should be 'manufacturers.id' in the resultSet.getObject() method.

src/main/java/mate/jdbc/dao/impl/CarDaoImpl.java
olekskov, line 50: The variable car in get method should be declared within the if block. It's not efficient to create a new object if it's not needed.

olekskov (approved):
Good job

